PR #291, opened 2016-11-19
Add auto_scaling

k2nr:
This PR adds functionality to automatically scales up/down ECS services.

## Example

```
  services:
    web:
      cpu: 128
      auto_scaling:
        min_count: 2
        max_count: 10
      # ...
```

A service scales by CPU metrics. Currently the configuration is as follows and this can't be customized.

- if Service's CPU average is between 40% and 60%, keep the current service scale
- if Service's CPU average is between 60% and 80%, add 10% of the current service scale
  - e.g. if the current scale is 2 and service's CPU average is 70%, auto scaling adds 1 to the current scale, thus it becomes 3 (2 * 10% == 0.2 and minimum scale addition is 1)
- if Service's CPU average is over 80%, adds 50% of the current scale
- If service's CPU average is under 40%, decreases 10% of the current scale
- The metric check happens every 3 minutes.
- By default both min and max counts are set to 2.

### DIFF
--- a/app/models/backend/ecs/v2/service_stack.rb
+++ b/app/models/backend/ecs/v2/service_stack.rb
@@ -43,7 +43,7 @@ module Backend::Ecs::V2
           build_alb_listener
         end
 
-        build_auto_scaling if service.auto_scaling.present?
+        build_auto_scaling
       end
 
       def build_service_role


k2nr:
Continues in #361

PR #315, opened 2016-12-31
Handle service CF stack failure

### LINKED ISSUE #300: Detect CloudFormation failure
Currently if service CF stack failed to update barcelonca just says it finished. the failure should be detected

k2nr:
Fixes #300


k2nr:
Continues in #362

PR #356, opened 2017-03-10
Add user option to oneoff

### LINKED ISSUE #345: Support user option for bcn run
It is a best practice to use non-root user for long running service. Currently oneoff tasks run with the default user that is specified in docker image, which is fine but when we run interactive oneoff via `bcn run` we sometimes want to do `apt-get` or some other commands that's require root permission. So it's very useful to do `bcn run` with `root` while long-running services use non-root user. 
Maybe it looks like `bcn run -e production --user root bash`

k2nr:
Fix #345

This supports `--user` option for `bcn run` e.g. `$ bcn run -e production --user root bash`

The `--user` option is already supported in the new client (which I will announce soon)

### DIFF
--- a/app/controllers/oneoffs_controller.rb
+++ b/app/controllers/oneoffs_controller.rb
@@ -29,7 +29,8 @@ class OneoffsController < ApplicationController
   def create_params
     params.permit(
       :command,
-      :memory
+      :memory,
+      :user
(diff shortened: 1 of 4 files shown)


k2nr:
Continues in #359

PR #359, opened 2017-03-13
Add user option to oneoff

### LINKED ISSUE #345: Support user option for bcn run

k2nr:
Fix #345

This supports `--user` option for `bcn run` e.g. `bcn run -e production --user root bash`

The `--user` option is already supported in the new client (which I will announce soon)

### DIFF
--- a/app/controllers/oneoffs_controller.rb
+++ b/app/controllers/oneoffs_controller.rb
@@ -29,7 +29,8 @@ class OneoffsController < ApplicationController
   def create_params
     params.permit(
       :command,
-      :memory
+      :memory,
+      :user
     )
   end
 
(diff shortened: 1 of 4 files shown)


degicat:
@shioyama Please review this.

k2nr:
This option is useful when you want to install a system package in the container e.g. `apt install vim`.

The docker best practice says container should be "unrooted" for example https://github.com/degica/barcelona/blob/master/Dockerfile#L7

If you unroot a container, the container user is no longer able to execute root command, which is great for security reason but for oneoff containers which die in a short period and don't open ports, it is not necessary. So this option allows you to do "unrooted by default, get root when you need" strategy

shioyama:
LGTM :+1:

k2nr:
Thanks!

PR #361, opened 2017-03-13
Add autoscaling

k2nr:


k2nr:
I'll add another scaling statistic: 95th percentile of ALB's response time. turned out that this metric is fairly easy to support

olingern:
@k2nr @rramsden Is this still being worked on / needed?

k2nr:
Yes I think so. It's not priority and I don't have time to work on this :sweat:

davidsiaw:
@k2nr I will be closing this PR since its really old and we are working on a new way to scale anyway

PR #362, opened 2017-03-13
Handle service CF stack failure

k2nr:


app/jobs/deploy_runner_job.rb
```diff
@@ -40,6 +40,6 @@ def perform(heritage, without_before_deploy:, description: "")
 
   def other_deploy_in_progress?(heritage)
     return false if heritage.version == 1
-    heritage.services.map { |s| !s.deployment_finished?(nil) }.any?
+    heritage.services.map { |s| s.deployment_status(nil) == :in_progress }.any?
```
olingern: Why pass `nil` here?

olingern:
This looks kind of straight forward, and looks like it would prevent deploying while another is in-progress.

Is there anything else that needs to be done to merge?

k2nr:
Looks like I just forgot to merge. I'll self-review again when I have time

PR #364, opened 2017-03-13
Tweak health check configuration

k2nr:
Changed as follows

- default health check interval 30s -> 10s
- healthy threshold 5 -> 2

The current configuration (30s * 5) makes deploy slower because it takes at least 2 mintues that a new container becomes healthy. with the new configuration deploy takes 10s at minimum


degicat:
@degicat Please review this.

degicat:
@rramsden Please review this.

k2nr:
@rramsden actually you don't need to review this. just a minor default value change and I confirmed that this works with AWS

PR #371, opened 2017-03-16
Send event when a user logs in and when a public key is signed

k2nr:
This PR adds very basic audit log feature. Looks like this

image


degicat:
@shioyama Please review this.

shioyama:
LGTM :+1:

k2nr:
Thanks!

PR #372, opened 2017-03-23
Install AWS Inspector agent and harden sshd config

### LINKED ISSUE #308: Install security inspection tool in container instances
PCI requirement (2.2)

AWS Inspector is the easiest option but it can be other OSS solution. I'll think about it

k2nr:
Fixes #308

This PR changes barcelona to install AWS Inspector agent when an instance launches. Currently Barcelona do not use Inspector but just installs the agent so that an infra administrator can run inspector against Barcelona instances without updating instances' userdata.

I also updated `/etc/ssh/sshd_config` to disable root login via SSH. this was found by Inspector when I ran it.

Ref: https://docs.aws.amazon.com/inspector/latest/userguide/inspector_working-with-agents.html

### DIFF
--- a/app/models/container_instance.rb
+++ b/app/models/container_instance.rb
@@ -12,6 +12,8 @@ class ContainerInstance
       "set -e",
       # Embed SHA2 hash dockercfg so that instance replacement happens when dockercfg is updated
       "# #{Digest::SHA256.hexdigest(district.dockercfg.to_s)}",
+
+      # Setup swap
       "MEMSIZE=`cat /proc/meminfo | grep MemTotal | awk '{print $2}'`",
       "if [ $MEMSIZE -lt 2097152 ]; then",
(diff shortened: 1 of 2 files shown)


k2nr:
So AWS Inspector agent is installed in all barcelona instances by default, which is I think fine because installing the agent doesn't cost money. It is totally opt-in feature; if an administrator wants to run inspector they can do it at any time otherwise Barcelona does nothing

degicat:
@rramsden Please review this.

k2nr:
@rramsden ping

rramsden:
Sorry missed this one. I'll take a look now.

k2nr:
Forgot to mention that this will fix #308

rramsden:
LGTM

k2nr:
Thanks!

PR #374, opened 2017-03-30
Remove NTP plugin and add NTP server to PCIDSS plugin

k2nr:
## Why
PCI DSS says:

> 10.4.1.a Examine the process for acquiring, distributing and storing the correct time within the organization to verify that:
> - Only the designated central time server(s) receives time signals from external sources, and time signals from external sources are based on International Atomic Time or UTC.
> - Where there is more than one designated time server, the time servers peer with one another to keep accurate time,
>- Systems receive time information only from designated central time server(s).

## How it works

image

- The plugin creates a new EC2 instance for NAT server. This server is managed by AutoScalingGroup (desired count == 1)
- The NAT server syncs its clock with public NTP servers
- When the NAT server launches, the server attach itself to ENI (Elastic Network Interface) that exists in the same AZ
- The plugin creates 2 ENIs in both AZ a and c
- Container instances refer both ENIs (one of the ENIs are not attached so doesn't respond)
- When the NAT server crashes for whatever reasons, AutoScalingGroup launch the same instance again (AZ is randomly chosen if both AZs are healthy)

## Notes

- Previously there was nat_plugin but I removed it and merged the feature to pcidss_plugin because in our case satisfying PCIDSS is the only case where we want to use this feature
- I tested in AWS environment, terminated NAT server several times to emulate hardware crash. It works fine: NAT server was re-launched by ASG and container instances recovered to sync clock in an hour
- There is another approach where NAT server registers its private IP to Route53 record `ntp.<district name>.bcn` and container instances refer that domain but this approach didn't work because `ntpd` resolves domain name when it launches and domain resolution never be refreshed after that.


degicat:
@camelmasa Please review this.

lib/barcelona/plugins/pcidss_plugin.rb
```diff
@@ -33,10 +305,29 @@ def on_network_stack_template(_stack, template)
 
         user_data = InstanceUserData.load_or_initialize(bastion_server["Properties"]["UserData"])
         user_data.packages += SYSTEM_PACKAGES
-        user_data.run_commands += RUN_COMMANDS
+        user_data.run_commands += run_commands
         bastion_server["Properties"]["UserData"] = user_data.build
         template
       end
+
+      def on_created(_, args)
```
camelmasa: Probably, you have forgotten to update `args` to `_` 😃

camelmasa:
LGTM 👍  Great 🎉

k2nr:
Tested AZ fail over again with modifyng container instance's clock to invalid time. NTP server re-launched correctly and container instance's clock was back to the correct time (it took 4 hours). Merging

PR #375, opened 2017-04-03
Limit bastion and ELBs' outbound ports

k2nr:
For ELB, I followed [this best practice](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-security-groups.html) with modification that all TCP ports are allowed for container instances.

### DIFF
--- a/lib/barcelona/network/vpc_builder.rb
+++ b/lib/barcelona/network/vpc_builder.rb
@@ -140,6 +140,14 @@ module Barcelona
           ]
         end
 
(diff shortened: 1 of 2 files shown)


k2nr:
@essa This works for basic HTTP/HTTPS web applications (both ALB and classic ELB) but can you review and check if this also works for apps that you manage?

essa:
I suppose this is related with the outbound traffic of ELBs and bastion servers, not with host instances.

If I'm right, it looks good to me. `bgproduction` heritage doesn't have nothing special about them.

k2nr:
Correct, nothing about instance configuration will change, just bastion and ELB. Thanks for your check.

PR #378, opened 2017-04-06
Better instance replacement 2nd step

k2nr:
This PR Changes the instance replacement strategy to https://aws.amazon.com/blogs/compute/how-to-automate-container-instance-draining-in-amazon-ecs/

This change should be applied in 2 steps

1. Add new resources (lifecycle hook and its lambda)
2. Remove the old script that checks current ECS status and increase rolling update batch size

This 2-step apply is needed because if we apply both changes at the same time, the lifecycle hook is not triggered (since it's being added and not ready) so ASG terminates old instances immediately

### DIFF
--- a/lib/barcelona/network/auto_scaling_group.rb
+++ b/lib/barcelona/network/auto_scaling_group.rb
@@ -11,6 +11,10 @@ module Barcelona
           j.DesiredCapacity desired_capacity
           j.Cooldown 0
           j.HealthCheckGracePeriod 0
+          # * 2: When instances are being replaced, instance count temporarily becomes
+          #      desired_count * 2
+          # + 1: There's a limitation "MinInstancesInService must be less than the autoscaling group's MaxSize"
+          #      without + 1, ASG cannot satisfy this requirement when desired_count == 0
           j.MaxSize(desired_capacity * 2 + 1)
           j.MinSize desired_capacity
           j.HealthCheckType "EC2"
--- a/lib/barcelona/network/auto_scaling_group.rb
+++ b/lib/barcelona/network/auto_scaling_group.rb
@@ -11,7 +11,7 @@ module Barcelona
           j.DesiredCapacity desired_capacity
           j.Cooldown 0
           j.HealthCheckGracePeriod 0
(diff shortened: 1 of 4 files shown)


k2nr:
@essa this PR is the second step. Can you review this? We can't merge this until we apply the latest CF template which includes the 1st step change

lib/barcelona/network/auto_scaling_group.rb
```diff
@@ -11,7 +11,7 @@ def define_resource(json)
           j.DesiredCapacity desired_capacity
           j.Cooldown 0
           j.HealthCheckGracePeriod 0
-          j.MaxSize(desired_capacity + 1)
+          j.MaxSize(desired_capacity * 2 + 1)
```
essa: Dose `MaxSize` here require `+1`?
I think this difference is for batch size of rolling update.

essa:
LGTM besides my comment.

lib/barcelona/network/auto_scaling_group.rb
k2nr: not batch size but `MinInstancesInService`. there is a limitation that is "MinInstancesInService must be less than the autoscaling group's MaxSize"

essa: Does `MinInstancesInService` become `desired_capacity * 2` on update?

k2nr: Sorry that was an answer to "why `+1` is needed". Here's why `* 2` is needed:
When rolling update runs, it doesn't change MaxSize of the autoscaling. What I want do in this PR is to replace all running instances at the same time so running instance count changes as follows

N (count before update) -> N*2 (old instances + new instances) -> N (new instances)

If `* 2` is not added, the above flow doesn't work because `MaxSize` is N + 1 and CloudFormation adds instances one by one

k2nr: Added a comment about it

essa: Thanks, I understand.

PR #380, opened 2017-04-06
Manage bastion servers as AutoScalingGroup

k2nr:
This PR makes bastion servers be managed by AutoScalingGroup. Previously a bastion server is launched as a plain "EC2::Instance" resource by CloudFormation. This way has some problems:

- If the bastion server crashes there's no way to recover it except manual way and if the bastion server is recovered manually, the server no longer belongs to the CloudFormation stack
- When we updates bastion server's UserData, CloudFormation only updates the running instance's UserData if a bastion server is managed as `EC2::Instance`. To run UserData's init script, we need to recreate the bastion server, which causes the first problem

Using AutoScalingGroup solves both problems. The name of "AutoScaling" is very confusing in this aspect; For bastion servers, ASG is not for "scaling" but for high availability and easy-to-replace capability.

### DIFF
--- a/app/models/district.rb
+++ b/app/models/district.rb
@@ -89,11 +89,14 @@ class District < ActiveRecord::Base
   end
 
   def bastion_ip
-    bastion_server_id = stack_resources["BastionServer"]
-    return nil if bastion_server_id.blank?
-
-    resp = aws.ec2.describe_instances(instance_ids: [bastion_server_id])
-    resp.reservations[0].instances[0].public_ip_address if resp.reservations.present?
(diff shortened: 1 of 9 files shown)


degicat:
@rramsden Please review this.

k2nr:
@rramsden ping. can you review when you have time?

rramsden:
Yikes 22 days ago, I'll take a look now

rramsden:
LGTM

k2nr:
rebased and updated AMIs. going to merge

PR #381, opened 2017-04-07
Better instance replacement 1st step: Add Lifecycle hook resources

k2nr:
As described in #378 this part of the change should be applied to the existing stack first.

### DIFF
--- a/drain_instance.py
+++ b/drain_instance.py
@@ -1,0 +1,43 @@
+import boto3
+import json
+import os
+import time
+
+session = boto3.session.Session()
+ecs = session.client(service_name='ecs')
+
+clusterName = os.environ["CLUSTER_NAME"]
+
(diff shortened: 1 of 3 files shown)


k2nr:
@essa This is the first step of #378 (cherry-picked the commit from that branch). Can you review?

drain_instance.py
```diff
@@ -0,0 +1,43 @@
+import boto3
```
k2nr: This lambda function is written in python because [the original blog post](https://aws.amazon.com/blogs/compute/how-to-automate-container-instance-draining-in-amazon-ecs/) uses python (I completely re-wrote the code though)

k2nr:
@essa ping. Can you review this?

essa:
LGTM 👍

k2nr:
Thanks!

PR #384, opened 2017-04-13
Healthcheck options

k2nr:
This PR adds missing health check options that are available ALB Target Group. See `barcelona.yml` for example

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -102,6 +102,9 @@ class HeritagesController < ApplicationController
             :endpoint,
             :health_check_interval,
             :health_check_path,
+            :health_check_timeout,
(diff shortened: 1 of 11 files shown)


k2nr:
@essa Can you review this? this is as you requested. The first 2 commits implement the feature, and next 2 commits are actually not relevant but I needed to implement to test the feature. You can see an example of how to use in `barcelona.yml`

olingern:
@essa is this close-able or is it still needed?

essa:
@olingern @k2nr 

I think you can close this for now.

I have had an issue but I resolved it by increasing reserved memory for the service. So it's not urgent for me.

Some legacy monolithic applications like Extinguisher can be unstable for health check because load for such services changes significantly. I think this feature may be useful for them but is not urgent.

PR #386, opened 2017-04-20
Add certificate identity to notifications

k2nr:
SSH sign notification will become like this `[default] Signed public key for user k2nr with identity k2nr@barcelona` which make it easier to know who is logging in what application


k2nr:
Just a notification change, going to merge

PR #387, opened 2017-04-21
Enable ELB logging

k2nr:
Added ELB loggin for both classic ELB and ALB
Ref https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html

### DIFF
--- a/lib/barcelona/network/vpc_builder.rb
+++ b/lib/barcelona/network/vpc_builder.rb
@@ -1,5 +1,23 @@
 module Barcelona
   module Network
+    # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html
+    ELB_ACCOUNT_IDS = {
+      "us-east-1"      => "127311923021",
+      "us-east-2"      => "033677994240",
+      "us-west-1"      => "027434742980",
+      "us-west-2"      => "797873946194",
+      "ca-central-1"   => "985666609251",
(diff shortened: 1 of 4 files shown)


lib/barcelona/network/vpc_builder.rb
```diff
@@ -375,6 +393,28 @@ def build_resources
...
+          j.Bucket stack.district.s3_bucket_name
+          j.PolicyDocument(
+            {
+              "Statement" => [
+                {
+                  "Action" => ["s3:PutObject"],
+                  "Effect" => "Allow",
+                  "Resource" => join("",
+                                     "arn:aws:s3:::",
+                                     "#{stack.district.s3_bucket_name}/elb_logs/*/AWSLogs/",
```
k2nr: This is a little different from the official manual. Since an S3 bucket can have only one policy, we need to define a bucket policy that enables all ELBs in a district to log.

k2nr:
@essa Could you review?

essa:
LGTM 👍

k2nr:
Thanks! I'll merge this now. After merging, I need to apply stack again to the existing districts so that all districts have the S3 bucket policy enabled.

To enable logging, we need to do `bcn endpoint update --district=<name> <endpoint-name>` for all endpoints.
For services that use  classic ELB, logging will be enabled at the next deploy

PR #394, opened 2017-05-10
Pass CA public key to remote

k2nr:
Without this, the created district in remote doesn't have CA public key so users can't ssh to the cluster


essa:
@k2nr 
Thanks for quick fix!

PR #396, opened 2017-05-15
Add/Use session token as a oneoff container identifier

k2nr:
Currently barcelona uses `oneoff.id` to identify an interactive container that `bcn run` connects to. This way has worked for a long time without any problem but this way actually has a potential security problem.

If someone of our team is evil and wants to hijack the oneoff interactive session that someone else launched, the evil person could do hijack by guessing interactive container ID (because, again, the current oneoff identifier is primary, auto-increment `id` which is guessable)

This PR makes the interactive oneoff hijack impossible by changing identifier from `id` to randomly-generated `session_token`. `session_token` is not stored in DB so only a person who creates an interactive oneoff can know the session token.

### DIFF
--- a/app/models/heritage_task_definition.rb
+++ b/app/models/heritage_task_definition.rb
@@ -21,7 +21,7 @@ class HeritageTaskDefinition
     new(heritage: oneoff.heritage,
         family_name: "#{oneoff.heritage.name}-oneoff",
         app_container_labels: {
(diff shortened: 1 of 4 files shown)


degicat:
@miyazawadegica Please review this.

k2nr:
Thanks!

PR #397, opened 2017-05-16
Bring Chaos to Barcelona

k2nr:
Hello Chaos!

This PR adds a very simple version of [Chaos Monkey](https://github.com/Netflix/SimianArmy/wiki/Chaos-Monkey).
This chaos script terminates oldest instances in each district.
The main purpose of this script is (I think) not to test fault tolerance, but to keep instances in Barcelona fresh so the termination strategy is not disruptive `ec2:TerminateInstances` but `autoscaling:TerminateInstanceInAutoScalingGroup` which triggers the autoscaling lifecycle hook to make instance termination graceful (ref: #378).

Also, this script doesn't include scheduling mechanism because we can use Barcelona's scheduled tasks

### DIFF
--- a/lib/barcelona/chaos.rb
+++ b/lib/barcelona/chaos.rb
@@ -1,0 +1,42 @@
+module Barcelona
+  class Chaos
+    attr_accessor :district, :terminate_count
+
+    def self.run(district_names, count:)
+      District.where(name: district_names.uniq).each do |d|
+        self.new(d, count).run
+      end
+    end
+
+    def initialize(district, terminate_count)
+      @district = district
+      @terminate_count = terminate_count
+    end
+
+    def run
+      instances = district.container_instances.sort_by { |i| i[:launch_time] }
+      non_active_count = instances.select { |i| i[:status] != 'ACTIVE' }.count
+      if non_active_count > 0
+        notify("#{district.name} has one or more non-active instance(s). Skipping", level: :warn)
+        return
+      end
+
+      asg = district.aws.autoscaling
+
+      # Terminate n oldest instances
+      terminate_count.times do |n|
+        ins = instances[n]
+        break if ins.nil?
+        notify "Terminating #{ins[:ec2_instance_id]}"
+        asg.terminate_instance_in_auto_scaling_group(
+          instance_id: ins[:ec2_instance_id],
+          should_decrement_desired_capacity: false
+        )
+      end
+    end
+
+    def notify(message, level: :good)
+      Event.new(district).notify(level: level, message: "[Chaos] #{message}")
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


k2nr:
Still WIP but I really think this script should be scheduled by default

essa:
I think this is a good idea but it will more useful if Barcelona set DRAINING to the instance before terminating it.

If Barcelona terminates instances immediately, we can't use this for services with `desired_count: 1`.

k2nr:
It does :smile: as I said in the description, this script calls autoscaling API instead of ec2:TerminateInstances so the draining lifecycle hook (the same one that is executed in rolling update) is triggered before actually killing an instance

k2nr:
I added chaos to barcelona's scheduled tasks. As an initial step, the scheduled task will only run for staging district.

degicat:
@shioyama Please review this.

lib/barcelona/chaos.rb
```diff
@@ -0,0 +1,42 @@
+module Barcelona
+  class Chaos
+    attr_accessor :district, :terminate_count
+
+    def self.run(district_names, count:)
+      District.where(name: district_names.uniq).each do |d|
+        self.new(d, count).run
```
shioyama: You don't need `self` here, just `new(d, count).run` should work.

lib/barcelona/chaos.rb
```diff
@@ -0,0 +1,42 @@
...
+      @terminate_count = terminate_count
+    end
+
+    def run
+      instances = district.container_instances.sort_by { |i| i[:launch_time] }
+      non_active_count = instances.select { |i| i[:status] != 'ACTIVE' }.count
+      if non_active_count > 0
+        notify("#{district.name} has one or more non-active instance(s). Skipping", level: :warn)
+        return
+      end
```
shioyama: Do you need a count here? If all you care about is that there is *any* match, think this should work:

```ruby
if instances.detect { |i| i[:status] != 'ACTIVE' }
  ...
end
```

spec/lib/barcelona/chaos_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      it "makes a terminate request for the instance" do
+        expect(asg_mock).to receive(:terminate_instance_in_auto_scaling_group).with(
+                              instance_id: 'i-11111111',
+                              should_decrement_desired_capacity: false
+                            )
+        chaos.run
+      end
+    end
```
shioyama: It might be worthwhile to test the case where there are multiple instances, and you only terminate a subset of them. A small code change could pass this spec but actually break that functionality.

shioyama:
LGTM :+1: Just a couple minor things

k2nr:
Thanks for your review! I'll merge it. It will start from tomorrow morning (staging only) let's see if it works :hocho:

shioyama:
Sounds good!

Issue #373, opened 2017-03-30
Install awslogs in bastion servers


k2nr:
Noticed awslogs is not installed in bastion servers..

PR #398, opened 2017-05-17
Security Fixes

k2nr:
This PR fixes some security issues about the network stack.

## Automate yum update

- install `yum-cron-security` package to instances except for container instances
- run `yum update --security` at launch for container instances

`yum-cron-security` periodically (daily) installs security updates if available. The reason why I didn't install the package to container instances is that updating packages of running instance might break instance functionality (for example, if there is an update for `docker`, it updates `docker` and **restarts** the daemon which makes all running containers temporarily down). Instead, I added `yum update --security` to container instances' user data. Now that we have chaos, running container instances periodically get terminated by chaos and new instances has latest security updates.

## Install AWS inspector in pcidss instances

PCIDSS plugin adds NTP server and OSSEC manager server, I forgot to install Inspector agent in those instances

## Install AWS logs agent in bastion server

Fixes https://github.com/degica/barcelona/issues/373

## Connect bastion server to OSSEC (PCIDSS plugin)

Bastion should be monitored by OSSEC

### DIFF
--- a/app/models/container_instance.rb
+++ b/app/models/container_instance.rb
@@ -12,6 +12,7 @@ class ContainerInstance
       "set -ex",
       # Embed SHA2 hash dockercfg so that instance replacement happens when dockercfg is updated
       "# #{Digest::SHA256.hexdigest(district.dockercfg.to_s)}",
+      "yum update -y --security",
 
       # Setup swap
       "MEMSIZE=`cat /proc/meminfo | grep MemTotal | awk '{print $2}'`",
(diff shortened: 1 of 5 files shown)


k2nr:
@essa I think you should review this. please take a look when you have time

essa:
LGTM.

I think this has no influence on bgproduciton district.

k2nr:
Found that bastion's available memory ran out because now bastion runs many services. I upgraded the instance type to `t2.micro` I'll see if `t2.micro` is enough for some days

k2nr:
Thanks for your review! I haven't seen any issue so far but I'll wait until tomorrow and if there's no issue I'll apply this change to all of our districts

PR #399, opened 2017-05-18
Enable Chaos for all districts

k2nr:
I confirmed that chaos gracefully terminated a staging instance this morning. Now I think it's ready to enable chaos for all districts

### DIFF
--- a/barcelona.yml
+++ b/barcelona.yml
@@ -5,7 +5,7 @@ environments:
     scheduled_tasks:
       # 10AM JST every week day
       - schedule: cron(0 1 ? * MON-FRI *)
-        command: bin/chaos --only staging
+        command: bin/chaos
     services:
       - name: web
         service_type: web


k2nr:
@essa Chaos will run for districts you manage too. Is it fine?

k2nr:
@essa ping

essa:
Well, I prefer longer interval but it's ok.

Only my concern is redis server for Sensu. Most of data on redis can be refreshed within one hour. But, if I disable some of alerts, it will be enabled on redis restart.

I don't disable alerts usually and only I have to do in such case is just disabling it again. So it's acceptable for me.

And I can move it to ElastiCache if I really keep them.

k2nr:
What interval do you think is the best? I was thinking other options and decided the currrent one (1 instance/day) because with this configuration **our** container instance lifetime is basically around 1 week which seems good length to me.

My other options were

- all instances / day
- all instances every 3 hour
- 1 instance every 1 hour

essa:
> our container instance lifetime is basically around 1 week which seems good length to me.

I agree with this current interval.

My usage of current redis is special. No need to consider about it.

k2nr:
OK thanks. If you come up with a better schedule please let me know :+1:

k2nr:
Chaos will run for all districts starting from next Monday

k2nr:
Chaos will run for all districts starting from next Monday

shioyama:
:+1:

PR #403, opened 2017-05-24
Fix PCIDSS plugin

k2nr:
- If there are 2 districts which both have PCIDSS plugin, then EBS to be attached may be conflicted so I added `barcelona` tag to the filter
- Add `OSSECManagerVolume` dependency on `OSSECManagerASG` to make sure that the volume is availabe when the manager volume starts
- Increase reserved memory from 128MB to 256MB. this is not necessary because container instances have swap area but in order to keep instances stable, it's time to increase the size since now contaienr isntances run many services outside ECS.


k2nr:
Those are small fixes. I'll merge and test on staging

PR #408, opened 2017-06-14
Add Vault auth

### LINKED ISSUE #401: Vault as an authentication/authorization backend
I was thinking about fine-grained authorization control https://github.com/degica/barcelona/issues/134 I came up with many ideas but all of them are not easy to implement. Authentication is relatively easy but authorization is always hard.

Instead of implementing access control inside Barcelona, now I'm thinking that it is easier and better to integrate HashiCorp's Vault and use it as a trusted external authorization service.

1. Every time user make a request to Barcelona, the request include Vault token
2. Barcelona calls Vault's `sys/capabilities-self` API with a path `secret/Barcelona/{{prefix}}/{{API path}}`
  1. `sys/capabilities-self` returns capablities on a given path for the token. For example, capabilities look like `["create", "read"]`
  2. Vault has built-in capabilities `create`, `update`, `read`, `delete`. In Barcelona, those capabilities are mapped to corresponding HTTP methods (e.g. `create` -> `POST`)
3. If the given token has a right permission for the API path and HTTP method, Barcelona allows executing the requested API

With this way Barcelona just don't need to care about authentication and authorization. Vault handles everything

## Example

Assume that a Vault token has the following capabilities

```
path "secret/Barcelona/prefix/v1/heritages/my-app*" {
  capabilities = ["read"]
}
```

users with this token can call get general information about `my-app`, can get environment variables of `my-app`, but cannot update/delete environment variables, cannot deploy, cannot do `bcn run` which calls `POST /v1/heritages/my-app/oneoffs`. And of course, this token do not have any permission on district API.

k2nr:
Fixes #401


k2nr:
@essa Can you review this? Majority of the change is refactoring, the commits that are related to Vault auth is f5678a8 and e9c582d

essa:
I think  the primary key of `User` should be auth_backend id + user name.

If two backend accidentally have same name for different person, and one uses Vault and the other uses github as auth backend, they can't use barcelona at a time.

This may be a very edge case which you need not to consider. But I think it's more natural because `User` recored is created by auth backend. So the identity of user is in the namespace of the backend.

k2nr:
Thanks. That is a nice catch, I'll update code so that the 2 auth types are not conflicted

k2nr:
@essa I added migration to add a new column `auth` to `users` table. I haven't changed the code to use the column, I'll do that in the next PR.
Do you think this is ready for merge?

essa:
LGTM 👍 
I have not found any issues besides my comment.

User recored will be accessed without index until next PR but it's not a problem for several users.

k2nr:
Thanks for your review! I'm going to merge this tomorrow :smile:

PR #410, opened 2017-06-22
Update EC2 AMI to the latest versions

k2nr:


k2nr:
I had to update pcidss plugin because updating the host OS version seems to have broken some components in pcidss plugin. I'll have some more tests and will merge tomorrow

k2nr:
Tested on AWS, the new AMIs worked fine

PR #412, opened 2017-07-05
Update ECS optimized AMI

k2nr:


k2nr:
Just updated AMI to the latest version. going to merge and test on staging

k2nr:
applied to staging. it's working normally (as far as I can tell) I'll keep watching

PR #414, opened 2017-07-19
Add OSSEC agent to NTP server

k2nr:
Adds OSSEC agent to the designated NTP server. code is getting very ugly, but I want to deploy this change and refactoring is the next task


rramsden:
LGTM

k2nr:
I had to open OSSEC outgoing ports. it's working, I'll have more tests and will merge tomorrow

PR #419, opened 2017-08-24
Do not alert syslog too large to slack

k2nr:
This is set as "noalert" here https://github.com/wazuh/wazuh-ruleset/blob/master/rules/0020-syslog_rules.xml#L30 but this setting doesn't stop sending alerts to slack which we use.
alerts whose level is higher than 12 is alerted to slack, so I set the level to `7`


k2nr:
@essa can you review?

essa:
LGTM 👍

k2nr:
Thanks! I'll merge and test on staging

PR #420, opened 2017-08-24
Update OSSEC docker configuration

k2nr:
I noticed the docker image build has been failing because of some upstream updates. I fixed it. I also made it to install specific version instead of the latest one.


k2nr:
No need to review, just changed minor versions and I'll test it on staging

PR #421, opened 2017-08-25
[WIP] Ignore long pending tasks on drainning

essa:
This is a temporary fix for an ECS issue which suddenly stops reporting task status.

If this issue happens, the tasks status remain 'PENDING' forever, even after completion. This PR enables terminating instance. It filters out long 'PENDING' tasks for count.


k2nr:
@essa Sorry I totally forgot to have a look at this. The current behavior is that AutoScalingGroup forcibly terminates instances after 1 hour wait. This is controlled by `HeartbeatTimeout` and `DefaultResult`. So I think we don't need to filter task status. What do you think?

essa:
@k2nr 
Yes, termination by ASG after one hour wait is enough.

This is needed when ECS is not reliable but it was fixed now. I will close this PR.

PR #422, opened 2017-09-01
Add ClamAV to NTP server

k2nr:
This PR adds ClamAV updater (freshclam) and daily ClamAV full scan to the dedicated NTP server. This configuration must be added but is missing in the current userdata script.

Once this is merged I'll apply this change to the clusters next week


k2nr:
@essa Could you review quickly?

essa:
LGTM 👍

k2nr:
Thanks!

PR #423, opened 2017-09-06
Update container instance AMI to 2017.03.f

k2nr:


k2nr:
Just Updated the AMI version. I'll merge and test on staging

k2nr:
I tested this on staging hoping it would fix the issue where sometimes container instances die but it made the situation worse... now sometimes docker daemon fails to start on launch :cry: I'm investigating

k2nr:
`Error starting daemon: error initializing graphdriver: devmapper: Could not query and save base device UUID:devmapper: Failed to find uuid for device /dev/mapper/docker-202:1-263631-base:signal: segmentation fault`

k2nr:
maybe the cause is `yum update --security` at the top of the user data `runcmd`. I removed it and the error didn't happen (I launched & terminated 30 times. still testing)

PR #424, opened 2017-09-07
Disable yum update --security on startup

k2nr:
I found that some times this setting causes weird errors so I'd like to disable it for now until we find a better solution


k2nr:
Note that Amazon Linux installs **critical** security patches on startup by default. The purpose of that setting was to install all security updates, not only critical ones

rramsden:
LGTM

PR #428, opened 2017-09-29
Resource based s3:GetObject ACL for container instances

k2nr:
Container instances must have minimum-required access to S3. Before merging this, we need to make sure that all of our services have ECS task role (services that have not deployed for a long time might not have a task role)


olingern:
@k2nr I think you can close this. S3 access is dependent upon policies for a given user. Unless this is for something different

k2nr:
[Container instances currently have full permission for "s3:GetObject"](https://github.com/degica/barcelona/blob/master/lib/barcelona/network/vpc_builder.rb#L326). This is very dangerous because credential data are stored in S3. there are only a few objects that container instances need to get. This PR is to minimize the container instance's get permission

davidsiaw:
I think with a little bit of checking that it works we can merge this.

PR #430, opened 2017-09-29
Refctor deploy runner job

k2nr:


olingern:
Auto-retries for *all* redeploys is maybe too liberal. There are some cases where AWS resources are in such a state that we should _not_ automatically redeploy. 

I've had issues with[ ECS agent connections dropping](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-agent-disconnected/) in the past, and the solution was to redeploy / kill the disconnected instance. I could see this retry logic being useful there, but not in all cases.

k2nr:
This PR is a pure refactoring meaning there's no behavior change. Barcelona doesn't retry deployments. If there's an error it just stops.

davidsiaw:
I fixed the conflict so I think this can already be merged.

degicat:
@essa please review this

PR #431, opened 2017-10-10
Update ECS optimized AMI

k2nr:
Updates the optimized AMI version to g


k2nr:
Just updated the AMI ID. will merge and test on the staging cluster

PR #432, opened 2017-11-08
Loosen lifecycle check period 1s -> 5s

k2nr:
checking container instance status every second seems to be too frequent so it may exceed the API rate limit. I changed the interval of 1 second to 5 seconds. 5 seconds in Lambda world is huge, but since the function is triggered only a few hundred times a day, I think it doesn't exceed the lambda free tier.


k2nr:
I also extended the timeout value of the function

PR #433, opened 2017-11-10
Update ECS optimized AMI -> 2017.09.b

k2nr:
This update includes the major version update of Amazon linux from 2017.03 to 2017.09. I'm still testing in the test cluster. I'll have some more tests and merge it next week if there's no issues


k2nr:
I'll merge and test on our staging cluster

PR #434, opened 2017-11-16
bundle update

k2nr:
Tests doesn't pass because of https://github.com/rails/rails/issues/30570 I'll wait for the bug to be fixed


k2nr:
Managed to make it pass by locking the `rack-test` version to `0.6.3`. I'm going to merge

PR #436, opened 2017-11-21
Set started_by to oneoff tasks

k2nr:
Sometimes oneoff interactive container remains running after a developer disconnected from it. When many of such "zombie containers" exist in a cluster, they occupies the entire cluster resource preventing new service containers from launching. So if it happens, I manually kill those zombie containers. If I can know "who created those oneoffs" it's easier to be sure if I can safely kill them

Below is the example of "started by". the empty one is the oneoff container.

image


k2nr:
@essa can you take a look quickly?

essa:
LGTM 👍 
It seems a good idea to me!

k2nr:
Thanks!

k2nr:
This is really nice :smile: 
image

PR #437, opened 2017-11-22
Update AMI to 2017.09.c

k2nr:
Release cycle is so frequent recently


k2nr:
Just an AMI version change as usual, I'm going to merge

PR #443, opened 2018-01-17
Use Amazon Time Sync Service instead of built-in ntpd

### LINKED ISSUE #441: Use Amazon Time Sync Service for instances time sync
https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/

I need to check if the service is PCIDSS compliant. If it is, we can remove the dedicated NTP server from PCIDSS plugin

k2nr:
Fixes #441 

The motivation is that with the time sync service, we don't need the complex NTP configuration in PCIDSS plugin (a dedicated NTP server, hot-standby ENI, right ENI selection)

the following error occurred during the initialization

```
+ service chronyd start
Generating chrony command key: tr: write error: Broken pipe
tr: write error
[  OK  ]
```

but I confirmed that the right chrony key is generated & stroed in the right place (/etc/chrony.keys)

the instance in test district with this configuration has been running for a few days and chrony correctly syncing time

### DIFF
--- a/app/models/container_instance.rb
+++ b/app/models/container_instance.rb
@@ -13,6 +13,10 @@ class ContainerInstance
       # Embed SHA2 hash dockercfg so that instance replacement happens when dockercfg is updated
       "# #{Digest::SHA256.hexdigest(district.dockercfg.to_s)}",
 
+      "yum erase -y ntp*",
+      "yum install -y chrony",
+      "service chronyd start",
+
       # Setup swap
       "MEMSIZE=`cat /proc/meminfo | grep MemTotal | awk '{print $2}'`",
       "if [ $MEMSIZE -lt 2097152 ]; then",
--- a/lib/barcelona/network/bastion_builder.rb
+++ b/lib/barcelona/network/bastion_builder.rb
@@ -168,6 +168,10 @@ module Barcelona
         EOS
 
         ud.run_commands += [
+          "yum erase -y ntp*",
+          "yum install -y chrony",
+          "service chronyd start",
+
           # awslogs
           "ec2_id=$(curl http://169.254.169.254/latest/meta-data/instance-id)",
           'sed -i -e "s/{ec2_id}/$ec2_id/g" /etc/awslogs/awslogs.conf',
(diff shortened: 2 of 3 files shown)


k2nr:
hard to debug because the error doesn't occur when I login via `ssh` and run `sudo service chronyd start` :disappointed:

k2nr:
I guess `/dev/urandom`'s entropy is not enough to generate the requested bytes :thinking:

k2nr:
OK although there is "broken pipe" error, correct key is appended to the key file. I'm not sure why the "broken pipe" happen but we can safely ignore it

k2nr:
Once this is merged and we confirm it works fine, I'll remove all code that are related to NTP. we no longer need the dedicated NTP server (PCIDSS plugin), security group (both inbound and outbound), network ACL, etc

k2nr:
Updated the description. @essa  can you review when you have time?

k2nr:
I just followed this doc https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html#configure-amazon-time-service

essa:
@k2nr 

I see no configuration for `chrony.conf`. 
Is the default of the chrony server in AWS (`169.254.169.123`)?

k2nr:
Amazon Linux's chrony package uses `169.254.169.123` by default. Here's the `chrony.conf` in my test district:

```
[ec2-user@ip-10-233-1-144 ~]$ sudo cat /etc/chrony.conf | head
# use the local instance NTP service, if available
server 169.254.169.123 prefer iburst

# Use public servers from the pool.ntp.org project.
# Please consider joining the pool (http://www.pool.ntp.org/join.html).
pool 2.amazon.pool.ntp.org iburst

# Record the rate at which the system clock gains/losses time.
driftfile /var/lib/chrony/drift
```

essa:
> Amazon Linux's chrony package uses 169.254.169.123 by default.
Great, thanks!

LGTM 👍

k2nr:
Thanks for the review! I'll keep the test cluster running until next Monday and will apply this to staging if there's no problem found

PR #444, opened 2018-01-23
Update ECS optimized AMI

k2nr:


k2nr:
Tested on the test cluster, worked fine

PR #464, opened 2018-07-03
Bundle update

k2nr:
- ran `bundle update`
- remove `rack-test` version lock
- Remove halt callback option
- Use FactoryBot


k2nr:
tests passing, I'll merge

PR #466, opened 2018-07-15
Bootstrap script requires secret key base

rramsden:
I noticed the bootstrap script no longer works with the current version of Rails. It's throwing an exception for me because it won't load the Rails environment without a secret key base defined.


k2nr:
Thanks!

PR #467, opened 2018-07-27
Update ECS optimized AMI to 2018.03.b

essa:
This PR will update the ami as following page.

* [Amazon ECS\-Optimized AMI \- Amazon Elastic Container Service](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html)

And it also adds a new region to `bin/show_ecs_images`. 

There is one more new region, "us-gov-west-1" but we can't `DescribeImage` of it. So I commented about it.

### DIFF
--- a/bin/show_ecs_images
+++ b/bin/show_ecs_images
@@ -7,6 +7,8 @@ unless IMAGE_LOCATION
 end
 
 AMAZON_OWNER_ID="591542846629"
+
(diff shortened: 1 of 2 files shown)


essa:
@k2nr 
Can you review this?

k2nr:
Thanks! LGTM :+1:

PR #471, opened 2018-08-16
Bump Rails to 5.2.1

greysteil:
Rails 5.2.1 is out, and is pretty good on backwards compatibility with 5.2.0, so this upgrade shouldn't cause any problems:

[Dependabot compatibility score](https://dependabot.com/compatibility-score.html?dependency-name=rails&package-manager=bundler&previous-version=5.2.0&new-version=5.2.1)

More generally, would you be up for using [Dependabot](https://dependabot.com) to keep the dependencies on Barcelona up-to-date? I ran it on my fork and it generated a [few PRs](https://github.com/greysteil/barcelona/pulls).

(I built Dependabot and would love to get you using it. It's totally free for open source and always will be - I just want to get more people using it so that the compatibility scores have more data. My plan is to then share that data with gem maintainers if/when there are any bugs in new releases, so that we can make the whole ecosystem more robust :octocat:.)


greysteil:
Sorry, hadn't noticed the test failure on this one. Just rebased and it's fixed.

k2nr:
Thank you for your contribution!

PR #478, opened 2018-10-16
Show amazon-linux ami as well as ecs-optimized

essa:
Resolves #477

This PR modifies `bin/show_ecs_image`

* Show the latest ami of amazon-linux as well as ecs-optimized
* Get the latest version of ecs-optimized
* Do it with `aws ssm` and public parameters 

Sample Output
```
$ bin/show_ecs_images
getting image id of /aws/service/ami-amazon-linux-latest/amzn-ami-hvm-x86_64-gp2...
        "us-east-1"      => ""ami-0ff8a91507f77f867"",
        "us-east-2"      => ""ami-0b59bfac6be064b78"",
        "us-west-1"      => ""ami-0bdb828fd58c52235"",
        "us-west-2"      => ""ami-a0cfeed8"",
        "eu-west-1"      => ""ami-047bb4163c506cd98"",
        "eu-west-2"      => ""ami-f976839e"",
        "eu-west-3"      => ""ami-0ebc281c20e89ba4b"",
        "eu-central-1"      => ""ami-0233214e13e500f77"",
        "ap-northeast-1"      => ""ami-06cd52961ce9f0d85"",
        "ap-northeast-2"      => ""ami-0a10b2721688ce9d2"",
        "ap-southeast-1"      => ""ami-08569b978cc4dfa10"",
        "ap-southeast-2"      => ""ami-09b42976632b27e9b"",
        "ca-central-1"      => ""ami-0b18956f"",
        "ap-south-1"      => ""ami-0912f71e06545ad88"",
        "sa-east-1"      => ""ami-07b14488da8ea02a0"",
getting image id of /aws/service/ecs/optimized-ami/amazon-linux/recommended ( latest version is amzn-ami-2018.03.g-amazon-ecs-optimized
 )...
        "us-east-1"      => "ami-0b9a214f40c38d5eb",
        "us-east-2"      => "ami-09a64272e7fe706b6",
        "us-west-1"      => "ami-0e7dd5fe55b87a5fe",
        "us-west-2"      => "ami-00430184c7bb49914",
        "eu-west-1"      => "ami-05b65c0f6a75c1c64",
        "eu-west-2"      => "ami-0209769f0c963e791",
        "eu-west-3"      => "ami-06b685336aa497c15",
        "eu-central-1"      => "ami-0bb804e8cd910a664",
        "ap-northeast-1"      => "ami-08681de00a0aae54f",
        "ap-northeast-2"      => "ami-0d947b1901b27a37c",
[...]

PR #479, opened 2018-10-23
Update AMIs for ecs instances, bastion instances and nat instances

essa:
I updated ami ids for ecs container instances, bastion instances and nat instances with [the new version of show_ecs_images]( https://github.com/degica/barcelona/pull/478).


essa:
@k2nr
Can you review this when you have time?

PR #487, opened 2018-12-03
Use bastion's AMI ID for ossec manager

k2nr:
Instead of hard-coding the ID here, it is better to refer the bastion's AMI ID since both instances should use the same AMI


k2nr:
@essa please review :bowing_man:

k2nr:
@essa ping

essa:
@k2nr 
Have you tested ossec manager on this AMI?

k2nr:
No, I want to test on staging after merge

essa:
Thanks, I see.

LGTM 👍

PR #488, opened 2018-12-03
bundle update

k2nr:


k2nr:
Tests passed, should be safe to merge

PR #492, opened 2019-02-12
Bundle update

k2nr:
Just updated gems with `bundle update`. Some tests failed due to [this change](https://github.com/thoughtbot/factory_bot/blob/master/NEWS#L3) so I fixed them


k2nr:
@essa can you review?

spec/factories/districts.rb
```diff
@@ -3,6 +3,7 @@
     sequence :name do |n|
       "district#{n}"
     end
+    region { 'us-east-1' }
```
essa: Why is this needed?

k2nr: As the factory_bot's changelog explains, the default strategy has been changed so that associations are by default "built" instead of "create". Previously associations were created and [this method https://github.com/degica/barcelona/blob/master/app/models/district.rb#L177] is called before validation. But now the method is not called since associations are just built.

The other spec change is made for the same reason.

essa: I see, thanks!

essa:
LGTM 👍

k2nr:
Thanks!

PR #494, opened 2019-02-21
Update Ruby version, Dockerfile, and Gemfile

k2nr:
- Ruby version 2.6
- Rewrote Dockerfile to use https://github.com/degica/dockerfiles/blob/master/rails-buildpack/2.6/Dockerfile and multi-stage build
- Install per-service aws-sdk gems instead of the all-in-one `aws-sdk`


Dockerfile
```diff
@@ -1,11 +1,38 @@
...
+      openssh-client \
+ && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /usr/local/bundle /usr/local/bundle
+ADD --chown=app:app . $APP_HOME
 
-ADD . $APP_HOME
-RUN chown -R app:app $APP_HOME
 USER app
-RUN rake assets:precompile
```
essa: Where and when are assets compiled?

k2nr: Well I've noticed Barcelona is API only and doesn't have any assets :sweat_smile: so I just removed this line

essa: Ah I see

essa:
LGTM 👍

k2nr:
Thanks!

PR #495, opened 2019-02-21
Support latest instance types for container instances

k2nr:
The regex looks crazy but couldn't find a better solution. we can remove that logic entirely when we drop support for instance types that do not support EBS-optimized (namely, t2 instances)


essa:
LGTM 👍

PR #497, opened 2019-02-21
Remove Network ACL

k2nr:
**Do not merge for now**

Removed Network ACL from the network stack. Currently we don't need this change but in the future we will this change to support VPC-mode task networking where ECS task uses its own ENI

### DIFF
--- a/lib/barcelona/network/subnet_builder.rb
+++ b/lib/barcelona/network/subnet_builder.rb
@@ -28,60 +28,6 @@ module Barcelona
           end
         end
 
-        add_resource("AWS::EC2::NetworkAcl", network_acl_name) do |j|
-          j.VpcId ref("VPC")
-          j.Tags [
-            tag("Name", join("-", cf_stack_name, network_type)),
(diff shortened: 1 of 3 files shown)


essa:
Why is Network ACL related with VPC-mode?

k2nr:
In bridge mode (which we currently use), each task has a bridge network interface that's connected to host's ENI, and host maps container port to host port, for example nginx container listens on port 80 and host maps it to host's 10000. In this case, port 10000 is the exposed port for the nginx container.
In VPC mode, each task has a dedicated ENI, so, if nginx container listens on port 80, that's ENI's port 80.

In bridge mode, we know that hosts do not listen on port ~ 1024 except for some special ports, which is why we filtered out most well-known ports with NACL. but in VPC mode, tasks can listen on any ports, some of which may need to be exposed to public subnets.

essa:
I see, thanks for detailed explanation.

LGTM 👍

davidsiaw:
@k2nr I'm gonna close this as well because its old and it doesn't look like we need it for now.

PR #498, opened 2019-02-21
Update ECS service role

k2nr:
This is to support ECS new features (DNS based service discovery, VPC-mode networking)


essa:
LGTM 👍

PR #499, opened 2019-02-21
Update bastion

k2nr:
- Use Amazon Linux 2 for bastion servers
- Removed unused ports from SG
- Added managed policies for SSM and CW logs and removed inline policy
- OSSEC manager do not start ssh daemon because we'll use SSM session manager


essa:
LGTM 👍

k2nr:
Updated the AMI again.
thanks for your review!

PR #500, opened 2019-02-21
container instance update

k2nr:
**Do not merge for now**

- Container instances now use Amazon Linux 2 based optimized AMI
- Installs SSM agent

### DIFF
--- a/app/models/container_instance.rb
+++ b/app/models/container_instance.rb
@@ -13,10 +13,6 @@ class ContainerInstance
       # Embed SHA2 hash dockercfg so that instance replacement happens when dockercfg is updated
       "# #{Digest::SHA256.hexdigest(district.dockercfg.to_s)}",
 
-      "yum erase -y ntp*",
-      "yum install -y chrony",
-      "service chronyd start",
-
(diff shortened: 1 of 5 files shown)


essa:
> Installs SSM agent

Is this mean that we will get rid of `bcn ssh` but continue to use `bcn run`?

k2nr:
You're right. Once we install ssm agent to all instances we can remove bcn ssh and we still need bastion servers to keep `bcn run` work

essa:
I see.

LGTM 👍

k2nr:
thanks. If you have an idea to get rid of bastion servers with SSM agent while keeping `bcn run` work, Please let me know.

Please don't merge this PR for now. I'd like to test Amazon Linux 2 with bastion servers first

k2nr:
I thought I would test AL2 with bastion servers but turned out that this PR needs to be merge in order to make Barcelona work due to this https://github.com/degica/barcelona/blob/master/lib/barcelona/plugins/pcidss_plugin.rb#L383

I'm going to merge this. Let's test on staging

PR #503, opened 2019-03-08
Allow ScheduleHandlerRole to PassRole

k2nr:
Found a bug where ScheduleHandler can't call `RunTask`. I added "TaskExecutionRole" to task definition so now exectuion of task is done in context of TaskExecutionRole.

I tested that this fixes the problem. You can see that now scheduled tasks are running correctly in "test-district" in our test AWS account


essa:
LGTM 👍

k2nr:
Thanks for your quick review!

PR #506, opened 2019-03-26
Domain environment variable

k2nr:
Sometimes it is useful that reviewapps know its domain name.


k2nr:
tests passing, I'll test this on staging

PR #511, opened 2019-04-10
bundle update

k2nr:


k2nr:
only minor updates and tests passed. going to :shipit:

PR #513, opened 2019-04-16
Preserve reviewapp listener config

k2nr:
Instead of entirely overriding `listeners`, Barcelona now add/update only necessary fields. With this change, the below config should work

```yaml
review:
  services:
    - name: web
      service_type: web
      # ...
      listeners:
        - health_check_path: /health_check
```

### DIFF
--- a/app/models/review_app.rb
+++ b/app/models/review_app.rb
@@ -15,13 +15,11 @@ class ReviewApp < ApplicationRecord
 
   def build_heritage
     web_service = services.find{ |s| s[:service_type].to_s == "web" }
-    web_service[:listeners] = [
-      {
-        endpoint: review_group.endpoint.name,
-        rule_priority: rule_priority_from_subject,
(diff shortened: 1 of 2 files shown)


k2nr:
@essa please have a look

essa:
LGTM 

But I think it would be better if [Matcher.HttpCode](https://github.com/degica/barcelona/blob/df6d8ebd72070f8faa0d761b27e8d1638b1b65ed/app/models/backend/ecs/v2/service_stack.rb#L197) is configurable.

This is needed for an app whose every service requires sign in like Basecamp. These app will redirect users to sign in page. So setting http code to 300-399 for health check may be a desired option for those apps.

This is not a change of review app but of heritage. So feel free to merge this PR as current version.

k2nr:
Thanks, I'll merge this now. Please open a new issue for the status code change

PR #515, opened 2019-06-14
Remove guard for declarative env

k2nr:
Secrets was added to CF :tada: I just removed the guard. This should work since reviewapps have been working but will have tests next week before merging

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -3,10 +3,6 @@ class HeritagesController < ApplicationController
   before_action :load_heritage, except: [:index, :create, :trigger]
   skip_before_action :authenticate, only: [:trigger]
 
-  # Declarative environment doesn't work for scheduled tasks because scheduled tasks use CloudFormation
-  # for creating task definitions and CF still does not support `ContainerDefinitions.Secrets` parameters.
-  before_action :forbid_environment, only: [:create, :update, :trigger], unless: -> { Rails.env.test? }
-
   def index
     render json: @district.heritages
   end
@@ -138,9 +134,5 @@ class HeritagesController < ApplicationController
 
   def load_district
     @district = District.find_by!(name: params[:district_id])
-  end
-
-  def forbid_environment
-    raise "Specifying environment is not yet supproted for normal deployments" if params[:environment].present?
   end
 end


essa:
@k2nr 
I'd like to use this feature, any problem with it?

k2nr:
No I think this will work. CF added support for ECS secrets and barcelona-cli already has `environments` feature for normal heritages.

I just didn't have time to test this. If you are interested, please go ahead and test this before merging this PR. This PR is not a breaking change (I think) but I want to make sure that this will work as expected before merge

k2nr:
I want to test the following with real AWS resources.

- Create new services
- Add a new environment entry
- Remove an existing environment entry
- Test combination of legacy env vars & new declarative environments
  1. create a service with legacy env A
  2. Add a new environment A (the same name as the legacy env) and confirm that the new A environment is used
  3. Remove the new A environment and observe that the legacy A appears
  4. Add new environment B and observe that both of the legacy A and the new B environment exist
  5. Remove the legacy A and observe that the service only has B environment

essa:
@k2nr 
I tested the combination of legacy env and new env by your scenario. It's working perfect! 

i. basecamp-web is running with a legacy env WORKER_PROCESSES=2 and I confirmed there are two unicorn-workers by ssh.
ii. I added following line to barcelona.yml and do `deploy -e staging --tag latest'

```yaml
  staging:
    name: basecamp2-staging
    ...
    environment:
      - name: WORKER_PROCESSES
        value: 1
```

And I confirmed there is only one unicorn-worker

iii. I remove the line and deployed again, then confirmed there are two workers.
iv. I added a new env with following barcelona.yml
```yaml
  staging:
    name: basecamp2-staging
    ...
    environment:
      - name: ENV_TEST
        value: ABC
 ```

Then I confirmed it by `bcn run -e staging env`

```
ENV_TEST=ABC
...
WORKER_PROCESS=2
...
```
v. I removed legacy WORKER_PROCESSES by `bcn env unset -H basecamp2-staging WORKER_PROCESSES`.

I have to set the number of tasks to zero because it is a required env.

Then I confirmed that it is removed and `ENV_TEST=ABC` remains by `bcn run -e staging env`.

k2nr:
Great! Thanks for testing :+1:

PR #517, opened 2019-07-31
Allow multiple listeners

k2nr:
https://aws.amazon.com/about-aws/whats-new/2019/07/amazon-ecs-services-now-support-multiple-load-balancer-target-groups/

I haven't tested this. I need to test before merging


k2nr:
This doesn't work as-is... There are some code that assume there is only one listener. I have to fix them

davidsiaw:
Closing because old wip

PR #519, opened 2019-10-07
Update nodejs runtime to nodejs8.10

k2nr:
`nodejs6.10` has been deprecated and that version can no longer be created or updated. In this PR I updated the version to `nodejs8.10` 

Note: The current latest runtime is `nodejs10.x` but CloudFormation still does not support 10.x with `ZipFile` option https://forums.aws.amazon.com/thread.jspa?messageID=908801

:heavy_check_mark: Tested on the test district on AWS


k2nr:
@essa Please review :bowing_man:

essa:
LGTM 👍

PR #523, opened 2019-10-18
Use libcurl3-gnutls instead of libcurl3 for buster

essa:
buster doesn't have `libcurl3` package and [the build failed on Quay](https://quay.io/repository/degica/barcelona/build/a8047d64-d450-4266-9df9-72214b41cd7f).

This fixes the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get update \
       libsqlite3-0 \
       libxslt1.1 \
       libxml2 \
-      libcurl3 \
+      libcurl3-gnutls \
       openssh-client \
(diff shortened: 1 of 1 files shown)


essa:
@k2nr 
Can you review this?

There is also `libcurl4` package with which the build succeeds. I couldn't find the difference of those packages so I selected the one which has the same major version number.

essa:
The build [succeeded](https://quay.io/repository/degica/barcelona/build/e8a5c990-f53a-4df1-b4b8-c82f1ae13a4e).

k2nr:
Thanks!

PR #527, opened 2019-11-15
New PCIDSS Plugins

essa:
This PR replaces pcidss_plugin with following three plugins.

#### ossec_client_plugin

This plugin installes wazuh client which connects to wazuh-manager running on provided host.

You need to give a hostname of wazuh-manager on activating it with following command.
```
$ bcn district put-plugin discrict1 ossec_client -a server_hostname=ossec-manager.local 
```

#### secure_instance_plugin

This plugin applies security updates before initializing and installs tools required PCIDSS standard, such as clamav and fail2ban.

#### itamae_plugin

This plugin installs [Itamae](https://github.com/itamae-kitchen/itamae) and download recipes from the specified location of S3 bucket.

You need to give a URL of the recipe archive on activating it with following command.
```
$ bcn district put-plugin essa-test itamae -a recipe_url=s3://barcelona-essa-test-1573458106/itamae_recipes/recipe.tar.gz
```

And it invokes itamae with following command on bastion instances.
```
$ itamae local roles/barcelona_bastion.rb -y nodes/barcelona_bastion.yml -l info
```

On container instances.
```
$ itamae local roles/barcelona_container.rb -y nodes/barcelona_container.yml -l info
```

So you must apply these role files and node files in the recipe.


essa:
@k2nr 
Can you review this?

lib/barcelona/plugins/itamae_plugin.rb
```diff
@@ -0,0 +1,73 @@
...
+      private
+      def install_itamae(user_data)
+        user_data.packages += %w(
+          gcc-c++
+          redhat-rpm-config
+        )
+        user_data.run_commands += <<~EOS.split("\n")
+          amazon-linux-extras install ruby2.4
+          yum install -y ruby-devel
+          gem install itamae io-console -N
```
k2nr: can you specify itamae version?

lib/barcelona/plugins/itamae_plugin.rb
```diff
@@ -0,0 +1,73 @@
...
+          mkdir /root/itamae && cd /root/itamae
+          aws s3 cp #{recipe_url} .
+          tar zxvf #{archive}
+          # Make sudo usable without tty
+          sed -i -e "s/^ *Defaults *requiretty/# Defaults requiretty/" /etc/sudoers
+          # apply itamae recipe
+          itamae local roles/#{role}.rb -y nodes/#{role}.yml -l info > /var/log/itamae.log
+        EOS
+
+        user_data.add_file("/etc/systemd/system/itamae.service", "root:root", "644", <<~EOS)
```
k2nr: Why is systemd used for executing itamae? is it not possible to just execute it inside the usedata?

lib/barcelona/plugins/itamae_plugin.rb
essa: Sure, I will update this.

lib/barcelona/plugins/itamae_plugin.rb
essa: There was an issue on redhat. Itamae uses `su` many times even if invoked on `root` and `su` fails on cloud-init on redhat. Invoking as a systemd service is a workaround for it.

This issue didn't happen on Amazon Linux2 but I used the same method for safety and uniformity. Because I couldn't figure out exact reason why `su` fails in such case.

I'm still wondering which to choose. How do you think about it?

lib/barcelona/plugins/secure_instance_plugin.rb
```diff
@@ -0,0 +1,81 @@
...
+        def self.extended(obj)
+          obj.configure_security
+        end
+
+        def configure_security
+          apply_security_update_on_the_first_boot
+          install_tools_for_pcidss
+        end
+
+        def apply_security_update_on_the_first_boot
```
k2nr: So container instances reboot on the first launch? I don't think we need this for container instances because Amazon releases security-patched AMI periodically

k2nr: basically I don't want to update existing packages for container instances because package updates might break compatibility and make instances stop functioning. For safety, We've chosen to just use plain ECS AMI and not to update yum packages.

k2nr: Actually we had `yum update --security` in container instances but removed that because there was a problem that caused ECS disconnection (If I remember correctly, dockerd died)

lib/barcelona/plugins/itamae_plugin.rb
k2nr: Hmm that is strange. I'd like to keep this script simple so if it works on amazon linux I want to choose the one-line solution.
Did the error occur every time on RedHat but it doesn't at all on Amazon Linux? If so, let's just try testing on staging several times to make sure that it is not an issue for Amazon Linux

essa: Thanks for the comment.

This issue happens only on RedHat, not on Amazon Linux2. So, I will remove the indirect invocation of Itamae.

Actually, `su` didn't fail but became very slow. I can find many similar issues in Internet.

* [Linuxのsuやsshが遅い \- 銀の鍵 \(The Silver Key\)](https://www.ginnokagi.com/2014/05/linuxsussh.html)

I think my issue is related with pam or name resolution on authentication as this post describes. But editing `/etc/hosts` didn't work.

I can watch the slow `su` while initialization continues. But `su` on my ssh was not slow at the same time. This means that I need to edit `/etc/hosts` or other configuration **before** `runcmd` starts to identify the exact cause.

So I abandoned investigation and used indirect invocation on RedHat.

lib/barcelona/plugins/secure_instance_plugin.rb
essa: I see, I will use AMI as is for container instances.

But I will update bastion instances because, while testing it, I have seen the newest AMI has high level vulnerabilities for a week until applied the security update.

I think we need a way to apply patches at least for instances in public network in case there are vulnerabilities in network stack of kernel.

k2nr: yes I agree

essa:
@k2nr
Can you review this again?

I have updated for your comments.

- Don't apply security update for container instances
- Apply Itamae directly without systemd service

And run AWS Inspector for bastion and container instances created by this plugin on local barcelona.

Then tested the results with a script which exclude trivial warnings.

Bastion instance has no warning.
```
$ aws-vault exec degica3 -- ruby check_inspector_result.rb cafes-test barcelona-essa-test-BastionAutoScaling-16MIG2I4L10L8
No important finding was found for barcelona-essa-test-BastionAutoScaling-16MIG2I4L10L8
```

Container instance got only warnings of updated packages.
```
$ aws-vault exec degica3 -- ruby check_inspector_result.rb cafes-test barcelona-essa-test-ContainerInstanceAutoScalingGroup-G99YZ3WC7374

CVE-2019-6470
rsyslog-0:8.24.0-16.amzn2.6.1
libseccomp-0:2.3.1-3.amzn2.0.3
CVE-2019-16056
```

lib/barcelona/plugins/ossec_client_plugin.rb
```diff
@@ -0,0 +1,56 @@
...
+# - Add VPC to Route53 Hosted Zone
+# - Create a peering connection
+# - Add an entry for peering to both Container VPC and ossec manager VPC
+# - Allow access from Container VPC to ossec manager VPC
+
+module Barcelona
+  module Plugins
+    class OssecClientPlugin < Base
+      def on_container_instance_user_data(_instance, user_data)
+        user_data.extend OssecClientUserData
```
k2nr: Why do you extend `user_data` here while in the itamae plugin it's not? I prefer to not extend the user_data object because the same object is used in the core so extending the object might cause a problem in the core
(I don't think there's any problem right now but maybe in the future we may add something that depends on modules that `user_data` extends or its class includes.)

lib/barcelona/plugins/secure_instance_plugin.rb
```diff
@@ -0,0 +1,80 @@
...
+
+module Barcelona
+  module Plugins
+   
+    class SecureInstancePlugin < Base
+
+      def on_container_instance_user_data(_instance, user_data)
+        user_data.extend SecureUserData
+        user_data.install_tools_for_pcidss
+        user_data.install_tcp_wrappers
```
k2nr: can you explain why is `tcp_wrappers` needed?

spec/lib/barcelona/plugins/ossec_client_plugin_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
```
k2nr: how about adding a spec to install all the plugins we use? Now that we have a bunch of plugins installed with some attributes, I'd like to know if there is no problem with the plugin combination we use.

lib/barcelona/plugins/secure_instance_plugin.rb
essa: Because a CIS rule requires it.

https://github.com/degica/barcelona/blob/abf9936a42ad943005a3f276746029ed0aa17989/spec/lib/barcelona/plugins/secure_instance_plugin_spec.rb#L37

k2nr: I see, so we don't install it in the bastion server because Inspector didn't complain?

lib/barcelona/plugins/ossec_client_plugin.rb
essa: I agree that consistency is important but I prefer separating UserData and plugin body by modules. So I'd like to re-write Itamae plugin in the same way as this one.

I think we should write new features as plugins from now on because we have important production districts and we don't want to break them on testing new features on staging district. Plugins can be enabled districts by districts

So we need a pattern for implementing coming plugins which may be more complicated. I'd like to recommend using modules and `extend` for following reasons.

1. module name and method name declares WHAT  while hiding HOW or implementation in method body
2. UserData should hide internals from its clients
3. some modules for UserData can be shared among many plugins
4. Later, we may need a callback with which Barcelona core calls plugin code at certain trigger, extending a module can be used naturally in such a case

I think UserData is an important object for plugins and the best way is encapsulate its internals by writing a class for it with a clear interface. But it will be a too big update and this pattern is a practical compromise

spec/lib/barcelona/plugins/ossec_client_plugin_spec.rb
essa: I think testing combination of plugins is a duty of integration tests. Bad configuration file or bad combination of packages can not be tested by unit tests anyway.

I have a plan to write a script or tool which tests a LaunchConfiguration. It will create a minimum ASG with a given LaunchConfiguration and an instance type. Then it verifies the EC2 instance launched on it.

And I think we can write it easily with AWS Step Function.

lib/barcelona/plugins/secure_instance_plugin.rb
essa: Yes, tcp_wrappers is already installed on bastion.

lib/barcelona/plugins/ossec_client_plugin.rb
k2nr: I see, then how about defining a module with a class method like this?

```ruby
  module OssecClientUserData
    def self.configure_ossec_client(user_data, server_hostname)
      # ...
    end
  end

  # And then call the method
  OssecClientUserData.configure_ossec_client(user_data, attributes["server_hostname"])
```

This way we don't need to make the user_data object dirty.
The only thing I'm worried about is method name conflicts. As long as we know method name is not going to conflict I think it's fine.
I know the methond name `configure_ossec_client` will be very rarely conflict, but if this pattern becomes a rule it's possible that somebody define a module with a common name like `configure`. This kind of bug can be very difficult to detect in development time because the issue only occurs with a specific plugin combination and order.

spec/lib/barcelona/plugins/ossec_client_plugin_spec.rb
k2nr: sounds good. thanks

PR #528, opened 2019-11-21
Add ssm_path to whitelist

essa:
Currently, Barcelona doesn't accept environment variables with `ssm_path`. 

When I post a `barcelona.yml` containing `ssm_path`, it fails with following message.

```
ActiveRecord::RecordInvalid: Validation failed: Environments value can't be blank, Environments value from can't be blank
```

This PR adds `ssm_path` to the whitelist for heritage requests.


essa:
@k2nr 
Can you review this?

k2nr:
Thanks! LGTM :+1:

PR #529, opened 2019-11-28
Set keepalive timeout to 75s

k2nr:
This PR changes nginx reverse proxy's `keepalive_timeout` to 75 seconds, which is [the default value](http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout)

The reason why this is needed is that nginx keepalive timeout should be longer than ALB's keepalive timeout, which is by default 60 seconds and we do not change it.
More detailed why is well explained here https://stackoverflow.com/a/50174356

### DIFF
--- a/dockerfiles/reverse-proxy/templates/alb_server.conf.erb
+++ b/dockerfiles/reverse-proxy/templates/alb_server.conf.erb
@@ -18,8 +18,6 @@ server {
   }
   <% end %>
 
-  keepalive_timeout 5;
-
   location / {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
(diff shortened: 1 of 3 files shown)


dockerfiles/reverse-proxy/templates/alb_server.conf.erb
```diff
@@ -18,8 +18,6 @@ server {
   }
   <% end %>
 
-  keepalive_timeout 5;
```
k2nr: `keepalive_timeout` in `server` block just overwrites the `http` config. we don't need to overwrite it so I just removed this line

k2nr:
@essa could you take a look?

essa:
LGTM 👍 

But I think we need to watch the behavior of shutting down or draining on deploying this. I'm worrying it might be longer because ECS services will not close connections until ALB closes.

k2nr:
Yes we definitely need to test before rolling this out. I think this doesn't affect deployment time since ALB initiates disconnection for idle connections when ALB needs to detach containers

davidsiaw:
This looks safe to deploy as-is for me can you confirm please @k2nr and @essa ?

PR #536, opened 2019-12-20
Update development docker files

k2nr:
The current docker files for development didn't work because excutable path resolution failed. I fixed the files so docker-compose work again

Previously `${PWD}` was the instance and `/app` was a symlink to $PWD. I reversed that and added `/app/bin` to `$PATH`


k2nr:
I'll merge this now

PR #537, opened 2020-01-14
Use ChangeSet for district update

### LINKED ISSUE #404: Use CloudFormation ChangeSet instead of direct update

k2nr:
Closes #404 

With this change, barcelona creates a [change set](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-changesets.html) instead of directly updating the stack.

I've been always worried a case where resources get deleted when we update a stack (`bcn district apply`). That can happen when we merge a wrong change into the master branch and the change contains a bug which deletes a resource unwillingly.

With this PR, Barcelona creates a change set so we can review what are going to change before we actually update a stack

### DIFF
--- a/app/services/apply_district.rb
+++ b/app/services/apply_district.rb
@@ -100,10 +100,6 @@ class ApplyDistrict
     aws.ecs.create_cluster(cluster_name: district.name)
   end
 
-  def create_or_update_network_stack
-    district.stack_executor.create_or_update
-  end
-
   def create_district_role(access_key_id, secret_access_key)
     task_role_arn = ecs_task_credentials&.dig("RoleArn")
     raise RuntimeError.new("Role ARN doesn't exist") if task_role_arn.nil?
--- a/app/services/apply_district.rb
+++ b/app/services/apply_district.rb
@@ -34,7 +34,7 @@ class ApplyDistrict
   def apply
     district.save!
     update_ecs_config
-    district.stack_executor.update
+    district.stack_executor.update(change_set: true)
   end
 
   def destroy!
--- a/lib/cloud_formation/executor.rb
+++ b/lib/cloud_formation/executor.rb
@@ -25,14 +25,27 @@ module CloudFormation
       client.create_stack(options)
     end
 
-    def update
-      client.update_stack(stack_options)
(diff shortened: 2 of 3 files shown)


app/services/apply_district.rb
```diff
@@ -100,10 +100,6 @@ def create_ecs_cluster
     aws.ecs.create_cluster(cluster_name: district.name)
   end
 
-  def create_or_update_network_stack
```
k2nr: not related to change set but I found that this method is not used so I just deleted it

k2nr:
@essa This is what I told to you today. can you review?

essa:
Can you tell me what I should do on following cases?

- (Case A): When I need to apply a urgent update on reviewing the previous change set which is not executed yet
  1. Revert the attribute of district 
  2. Update the attribute of district for the urgent change
  3. Invoke `bcn district apply` and execute the urgent change set
     - the first changeset is disabled automatically by AWS, right?
  4. Update the attribute of district for the first change again
  5. Invoke `bcn district apply` and execute the change set
- (Case B): When a change set failed on execution because of an out of stack issue
  1. The change set is disabled automatically on failure by AWS, right?
  2. Fix the issue
  3. Invoke  `bcn district apply` 
  4. Execute it ( I  can see only the latest change on AWS Console )

What I'm wondering is whether it works on edge case when there is a changeset which is not completed normally.

k2nr:
ChangeSet behaves as following

- You can create multiple changesets for one stack. Until one of those is executed, all changesets are active; think change set as a pull request in GitHub
- When one change set is executed, other changesets for the same stack get deleted

For Case A, it works as you described. the first changeset gets deleted at the step 3.

For Case B, once changeset is "executed" all changesets get deleted. So

- the first change set is deleted when it is executed. it doesn't matter whether stack update completes or not
- at the step 4, the change set shows diff between "before step 1" and the latest change (that is contained in the changeset) because the stack is now "UPDATE_ROLLBACK_COMPLETE" status.

essa:
Thanks for the explanation! I think I understood.

LGTM 👍

k2nr:
Thanks!

PR #540, opened 2020-01-17
Allow connecting to external ossec servers

essa:
This PR enables bastion severs to connect to ossec servers located outside of the VPC.


davidsiaw:
Is this ready to be reviewed @essa?

essa:
@davidsiaw 
This will work but, strictly thinking, it is against PCIDSS compliance because we need to allow OSSEC connection only to ossec server.
I have updated the SG manually so that we allow minimum cidr block.

I think this PR should be closed for now and rewritten as a part of ossec plugin.

davidsiaw:
That makes sense @essa. Thank you. I will close this PR then.

PR #541, opened 2020-01-22
Fix secure instance plugin

essa:
This PR fixes secure instance plugin.

Rebooting in the `bootcmd` of cloud-init seems to have a flipflop issue. 

I tested it by booting bastion instance in a district only with this plugin several times. Although the launch configuration was not changed, the instance became silent with no message some times.

So I moved the script in the `bootcmd` to the beginning of the `runcmd` and make it run on every boot with `cloud_final_modules: [['scripts-user', 'always']]`.

### Tricky part of this PR

I need to guarantee that rebooting is the first command in the `runcmd` so that any commands will run just one time.

Currently I can only put it to the first of `run_commands` array.
```ruby
self.run_commands.unshift *<<~EOS.split("\n")
...
EOS
```
This works when this plugin is the only plugin which puts commands with `unshift`. I think it's a little bit tricky but adding an interface for it to `InstanceUserData` seemed too much to me.

### DIFF
--- a/app/models/instance_user_data.rb
+++ b/app/models/instance_user_data.rb
@@ -1,5 +1,5 @@
 class InstanceUserData
-  attr_accessor :files, :boot_commands, :run_commands, :packages, :users
+  attr_accessor :files, :boot_commands, :run_commands, :packages, :users, :cloud_final_modules
 
   def self.load_or_initialize(base64 = nil)
     if base64.nil?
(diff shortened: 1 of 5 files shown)


essa:
@k2nr 
Can you review this?

k2nr:
LGTM :+1: 

> This works when this plugin is the only plugin which puts commands with unshift

I think this is an acceptable limitation.

PR #545, opened 2020-02-13
Permissions framework for barcelona

### LINKED ISSUE #546: Fine-grained permission control for Barcelona
Currently all users who have access to Barcelona are either admins or developers. The lesser access level, developers while only barred from creation and deletion of districts, still have large and wide-ranging permissions across all projects.

We need a better way to determine per-project per-command permissions better.

davidsiaw:
# Context

This is a permissions framework that is both detailed and flexible. The relationship between an action and permission is simple. For example if a person wishes to allow developers to create districts then the permission is `district.create`. If a person wishes to allow developers to edit a district called `sagrada` the permission is `district.edit.sagrada`.

This opens up the possibility of giving/removing wide ranging permissions by simply iterating through all things. To give permissions to a developer `alice` to edit all districts we simply iterate through all districts to create the `Permission.create(user: alice, key: "district.edit.#{district.name}")` for example. Vice versa to remove all.

Users only have one permission by default at the beginning `users.edit` which can also be revoked, leaving the user unable to change his own ssh keys.

Admins have full permissions regardless of what permissions they were given before.

fixes #546 

This PR really is just here to implement the minimum viable product for permissions. There are other desirable features that can be added to it over time.

## What it does not have right now
- API for admins/those with permission to add permissions
- Wildcard permissions?
- API to add/remove ranges of permissions (allow all developers to create endpoints will iterate through all the developers and give them the permission `endpoint.create` for example)
- Endpoints permissions
- Not sure how to `bcn run`

# How to test
1. Login as a github user (non-admin)
2. Attempt to perform `bcn district list`
3. Observe that it fails
4. Use `rails c` to create a permission `Permission.create(user: User.last, key: 'district.list')`
5. Attempt `bcn district list`
6. Observe that it succeeds

### DIFF
--- a/Kaiserfile
+++ b/Kaiserfile
@@ -1,4 +1,8 @@
-dockerfile 'Dockerfile'
+require 'fileutils'
+
+File.write('.kaiser.dockerfile', File.read('Dockerfile').sub('--without development test', ''))
+dockerfile '.kaiser.dockerfile'
+FileUtils.rm '.kaiser.dockerfile'
 
 db 'postgres:alpine',
    port: 5432,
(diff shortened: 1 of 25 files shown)


davidsiaw:
Closing this down since stale

PR #552, opened 2020-02-26
Fix ACM notes

### LINKED ISSUE #377: Bootstrap script does not create HTTPs ELB when ACM cert is not provided
Hi there, (me again)

I am trying out Barcelona by following getting_started.md .

I found that if I did not specify an ACM Certificate during bootstrap, Barcelona won't create an ELB listener for HTTPS, and:

* HTTPS could not be connected
  * It is not possible to do `bcn login https://blah-blah` because 443 is not enabled

I understand that it's not possible to create a HTTPS listener on ELB without a certificate. However, I think it's worth to add more description to the bootstrap message:

```
Create a new certificate on Amazon Certificate Manager in the region you chose and provide its ARN.
This is optional but we highly recommend to setup ACM certificate.
```

To something that asks the user to enable HTTPS listener itself, after the provisioning is finished.

Thanks!

davidsiaw:
This PR just makes it really obvious that setting up an ACM is really important for having a barcelona district in the docs

Fixes #377

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -39,7 +39,9 @@ EOS
     ENV["ACM_CERT_ARN"] = options[:acm_cert_arn].presence ||
                           ask_with_description("ACM Certificate ARN", <<-EOS, required: false)
 Create a new certificate on Amazon Certificate Manager in the region you chose and provide its ARN.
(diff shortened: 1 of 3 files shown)


degicat:
@Resonious please review this

PR #555, opened 2020-02-27
Make it possible for ALB to accept PortMappings in barcelona.yml

### LINKED ISSUE #553: Portmapping stuck to pointing at container:3000
# Context

For some reason even when

```ruby
    services:
        port_mappings:
          - lb_port: 80
            container_port: 8080
```

is set, the container port used continues to be set to port 3000.

スクリーンショット 2020-02-27 11 09 34

# How to test
You will need a district
1. Check out `git clone https://github.com/degica/barcelona-hello`
2. `cd barcelona-hello`
3. Create a barcelona endpoint `bcn endpoint create --public --district=<name> hello2`
4. Create a service `bcn create -e hello -d <name>`
5. Log in to aws and notice that the task definition points to port 3000 on the container

davidsiaw:
This PR removes the hardcoded port 3000 for the webserver.

Fixes #553 

## How to test
Try the thing in #553 and observe that it reaches the webserver.

Also open the AWS console and observe the ports are set to the ones selected in barcelona.yml.

### DIFF
--- a/app/models/heritage_task_definition.rb
+++ b/app/models/heritage_task_definition.rb
@@ -13,7 +13,8 @@ class HeritageTaskDefinition
         force_ssl: service.force_ssl,
         hosts: service.hosts,
         reverse_proxy_image: service.reverse_proxy_image,
-        mode: (service.listeners.present?) ? :alb : :tcp
+        mode: (service.listeners.present?) ? :alb : :tcp,
+        web_container_port: service.web_container_port
        )
   end
 
@@ -73,7 +74,7 @@ class HeritageTaskDefinition
     end
   end
 
-  def initialize(heritage:, family_name:, user: nil, cpu:, memory:, command: nil, port_mappings: nil, is_web_service: false, force_ssl: false, hosts: [], app_container_labels: {}, reverse_proxy_image: nil, mode: nil)
+  def initialize(heritage:, family_name:, user: nil, cpu:, memory:, command: nil, port_mappings: nil, is_web_service: false, force_ssl: false, hosts: [], app_container_labels: {}, reverse_proxy_image: nil, mode: nil, web_container_port: 3000)
     @heritage = heritage
     @family_name = family_name
     @user = user
@@ -87,6 +88,7 @@ class HeritageTaskDefinition
     @reverse_proxy_image = reverse_proxy_image
     @app_container_labels = app_container_labels
     @mode = mode
+    @web_container_port = web_container_port
   end
 
   def web_service?
@@ -94,7 +96,7 @@ class HeritageTaskDefinition
   end
 
   def web_container_port
-    3000
+    @web_container_port
   end
 
   def container_definition
--- a/app/models/port_mapping.rb
+++ b/app/models/port_mapping.rb
@@ -22,6 +22,7 @@ class PortMapping < ActiveRecord::Base
   scope :lb_registerable, -> { where.not(protocol: "udp") }
   scope :tcp, -> { where(protocol: "tcp") }
   scope :udp, -> { where(protocol: "udp") }
+  scope :http, -> { where(protocol: "http") }
 
   def self.to_task_definition
     self.all.map { |pm|
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -93,10 +93,14 @@ class Service < ActiveRecord::Base
 
   def create_port_mappings
     return unless web?
-    return if self.port_mappings.count.nonzero? # No need to create these if already specified
 
-    self.port_mappings.create!(container_port: web_container_port, protocol: 'http')
-    self.port_mappings.create!(container_port: web_container_port, protocol: 'https')
+    if http_port_mapping.blank?
+      self.port_mappings.create!(container_port: web_container_port, protocol: 'http')
+    end
+
+    if https_port_mapping.blank?
+      self.port_mappings.create!(container_port: web_container_port, protocol: 'https')
+    end
   end
 
   def backend
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -63,7 +63,9 @@ class Service < ActiveRecord::Base
   end
 
   def web_container_port
-    3000
+    return 3000 unless http_port_mapping.present?
+
+    http_port_mapping.container_port
   end
 
   def http_port_mapping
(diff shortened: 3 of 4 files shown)


degicat:
@k2nr please review this

degicat:
@essa please review this

essa:
@davidsiaw 
I tested it with my local barcelona.

It worked for a new created service but it failed on updating the port mappings on following two cases.

1. Create a service without port mappings and update it with a `barcelona.yml` with portmappings
2. Create a service with port mappings as you specified and update the container port

Both cases preserve the port mappings which has been created on initialize although `barcelona.yml' was updated.

@k2nr 
I think we need to destroy all the port mappings on update. How do you think?

davidsiaw:
Ah, so the error is in barcelona because duplicate HTTP ports appear right?

Yes I think the automatically created port mappings need to be removed and readded on change. I think services should be explicit on what ports they have because they are behind a reverse proxy and they do not need to handle HTTPS.

This might be old code that we no longer need. If we should preserve compatibility I am still not sure how to do that properly

essa:
Or you can introduce a new attribute of service and don't use `port_mappings` for this use case.

```yaml
environments:
  hello:
    name: hello
    image_name: quay.io/dsiawdegica/hello
    scheduled_tasks: []
    services:
      - name: web
        service_type: web
        memory: 256
        command: sh /start.sh
        container_port: 8080 # <- new attribute
        listeners:
          - endpoint: hello
            health_check_path: /
```

davidsiaw:
I actually had an idea, instead of adding the portmappings on create, we can just add it when creating the task definition, this way they are only there if no other portmappings are present, preserving compatibility. I will do this and try your failure cases

essa:
@davidsiaw 

> adding portmappings to the template, but not create any portmapping objects

I partly agree with you because services are created or updated always with `barcelona.yml`. So detailed information of services needs not to be stored in DB.

But two considerations are required for this idea.

1. How to provide information about port mappings for `show heritage` API
2. We can't move the method which uses this attribute to delayed jobs ( in the future )

So I think it can be a big change of architecture of Barcelona and it should be discussed in a long term with other attributes of services.

davidsiaw:
> How to provide information about port mappings for show heritage API

We don't show it anyway. I think if we wanted to add it that is a separate concern

```
$ ./bcn app show hello
Name:          hello
Image Name:    quay.io/dsiawdegica/hello
Image Tag :    latest
Version:       2
Before Deploy: None
Token:         864ff853-4413-4e2e-9ed4-19493fcc70d9
Scheduled Tasks:
Environment Variables
```

> We can't move the method which uses this attribute to delayed jobs ( in the future )

We still can because the code that adds it to the template has nothing to do with delayed jobs. Specifically it will be the `container_definition` method in the `heritage_task_definition.rb` file. It will be the same code that will be run anyway in delayed jobs.

essa:
> We don't show it anyway. 

No, barcelona has an api which shows it.

```
$ bcn api get /heritages/jet-jaguar-staging
...
 "services": [
 ...
 {
        "command": "unicorn -c config/unicorn.rb",
        "cpu": 128,
        "desired_count": 2,
        "endpoint": null,
        "force_ssl": true,
        "health_check": {
          "protocol": "tcp"
        },
        "hosts": [],
        "memory": 512,
        "name": "web",
        "pending_count": 0,
        "port_mappings": [
          {
            "container_port": 3000,
            "host_port": 12405,
            "lb_port": 80,
            "protocol": "http"
          },
          {
            "container_port": 3000,
            "host_port": 19830,
            "lb_port": 443,
            "protocol": "https"
          }
        ],
```

I think Barcelona has a great integrity in its architecture. This API is a sample of it.

We can create a tool around it easily because everything is transparent through REST APIs. This architecture can keep the core of Barcelona simple and small if the number of users grows.

I don't say you must not change it but we need to distinguish changes **inside of** the current architecture and changes **of** the architecture itself.

davidsiaw:
Is it important that we keep it? There are two solutions to this:

1. We abstract away the portmapping property to add the default portmapping if not provided

2. We do not show it if the user never provided it in his barcelona.yml. Instead we add an API to inspect the AWS properties directly.

essa:
> Is it important that we keep it?

It's a trivial thing but it's important to treat all trivial things in a uniform way.

Because consistency saves cost of  reading codes. You can imagine or expect behaviors and implementation without reading codes or documents if all the trivial things are treated in a uniform way.

Barcelona is REST API for its own resources with event handlers. It's simple because half of it is a typical basic Rails Application and rest of them is event handlers for resources. 

For example, you can separate problems on most of troubles into two parts, "Does the Barcelona resource is what I expected?" and "Why the resource is not applied to the AWS resources?". This is a great virtue of consistency.

davidsiaw:
@essa thanks for your feedback. Do you think we should keep the automatic portmapping then? I think this is a bigger discussion about the way barcelona handles changes to the stack for two kinds of load balancers: the ALB and the ELB.

My issue is right now it is impossible with barcelona to deploy a container that serves on a port that is not 3000. I have other ideas on how to fix it but I am not sure what the best way is

davidsiaw:
@k2nr @essa I'm not sure how to solve this problem perfectly.

Because of this bug It is impossible with barcelona to deploy a container that serves on a port that is not 3000.

This is just one solution, and I think we should be able to move forward with this and consider refinements later

essa:
@davidsiaw 

Sorry for late response.
If you hurry, please use the simplest solution: adding a new attribute of `Service`, `web_container_port`.

```
environments:
  hello:
    name: hello
    image_name: quay.io/dsiawdegica/hello
    scheduled_tasks: []
    services:
      - name: web
        service_type: web
        memory: 256
        command: sh /start.sh
        web_container_port: 8080 # <- new attribute
        listeners:
          - endpoint: hello
            health_check_path: /
```

This is simple because it doesn't require thinking about using `port_mappings` for `type: web` services.

This attribute already exists as a method which returns constant.

```ruby
def web_container_port
    3000
end
```

So all you need is adding a new field to the `services` table and removing this method.

Your solution using `port_mappings` with `type: web` service is flexible but requires much consideration. We may suffer compatibility issues later. I can't approve it without long discussion.

davidsiaw:
This was solved in https://github.com/degica/barcelona/pull/583

PR #561, opened 2020-03-09
Remove CPU defaults for barcelona

### LINKED ISSUE #558: Make CPU default to nil instead of 512 on barcelona
Seems weird that CPU is set to 512 even though cloudformation task definition says it is completely optional.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html

davidsiaw:
# Context

This removes the default of 256 cpu units when CPU is not specified during the creation of a heritage. The issue linked (#558) describes a value of 512 which was a default that was specified by the barcelona cli. This PR removes the server's default to omit mentioning it if it is not set in b arcelona.yml.

Fixes #558

### DIFF
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -271,7 +271,6 @@ class Heritage < ActiveRecord::Base
   def base_task_definition(task_name, with_environment: true)
     base = district.base_task_definition.merge(
       name: task_name,
-      cpu: 256,
       memory: 256,
       essential: true,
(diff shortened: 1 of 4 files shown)


k2nr:
Is this ready for review? the change LGTM

davidsiaw:
Yes sorry. I thought I put it up for review but I didn't

PR #564, opened 2020-03-12
Fix CPU not defaulting to nil when erasing the member from barcelona.yml

### LINKED ISSUE #563: Deleting barcelona.yml cpu does not modify existing application
Not sure if this is a legitimate issue, but just marking it as a known one.

If a heritage already exists, and you modify the cpu setting of the application, a newly created task definition will still have the old setting.

davidsiaw:
# Context

As in the ticket, this PR fixes CPU not being set to nil when removed from `barcelona.yml`.

Fixes #563 

This change may be controversial as the original specs mandated that a PATCH only changes attributes that were mentioned. This change breaks that behavior for the `service.cpu` attribute specifically by setting it to `nil` when it is not mentioned.

This is possibly a case of use-case does not match specification. The current normal use-case is to have a `barcelona.yml` read in by `barcelona-cli` and sent to `barcelona` using a PATCH method.

This way of fixing this is admittedly hacky. An alternate solution, that I just thought of while writing this post was to make it so we have a `replace` action that maps to the PUT method that basically always takes a fully-formed definition instead of a partial one  that we illustrate in our tests.

### DIFF
--- a/app/services/build_heritage.rb
+++ b/app/services/build_heritage.rb
@@ -91,8 +91,8 @@ class BuildHeritage
 
   def execute
     heritage.district = district if district.present?
-    heritage.attributes = prenullify_params
-    heritage.attributes = params
+    heritage.assign_attributes prenullify_params
+    heritage.assign_attributes params
     heritage
   end
 end
--- a/app/services/build_heritage.rb
+++ b/app/services/build_heritage.rb
@@ -77,6 +77,9 @@ class BuildHeritage
     [:cpu]
   end
 
+  # This method sets fields defined in #nullables to nil before applying
+  # updates to them, so that any deletions in barcelona.yml are detected
+  # properly.
   def prenullify_params
     nullifiers = []
     heritage.services.each do |service|
--- a/app/services/build_heritage.rb
+++ b/app/services/build_heritage.rb
@@ -34,9 +34,10 @@ class BuildHeritage
       new_params[:environments_attributes] = new_params.delete(:environment).map do |e|
         e.slice(:name, :value, :value_from, :ssm_path)
       end
-      new_params[:environments_attributes] = sync_resources(new_params[:environments_attributes], heritage.environments, :name)
+      new_params[:environments_attributes] = sync_resources(
+        new_params[:environments_attributes], heritage.environments, :name
+      )
     end
-
 
     unless heritage.new_record?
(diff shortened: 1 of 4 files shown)


davidsiaw:
@k2nr and @essa you might want to have a look at this. It was really confusing why this was happening but it turns out we don't update the service completely and the code that marshals the data into a form usable by the rails mass assignment is fairly complex.

I finally figured out a way to make this work after a day of debugging but I think it may be a bit too hacky.

I wrote down an alternate solution in the opening post but please let me know what you think of this solution

app/services/build_heritage.rb
```diff
@@ -57,22 +58,40 @@ def convert_params_for_model(new_params)
...
+      nullifiers << { id: service.id, **nuller_hash }
+    end
+    {"services_attributes" => nullifiers}
+  end
+
   def execute
     heritage.district = district if district.present?
+    heritage.attributes = prenullify_params
     heritage.attributes = params
     heritage
```
essa: `prenullify_params` seems to be replaced by `params`, what's the effect of `prenullify_params`?

davidsiaw: This actually performs a rails mass assignment. I expected this to be confusing. This `=` operator takes a hash and then assigns values into the attributes of the given object. See https://apidock.com/rails/ActiveModel/AttributeAssignment/attributes%3D

I will use the `assign_attributes` method instead to make this clearer. Thanks for pointing it out.

davidsiaw:
@essa I finally fixed the tests and changed it to use `assign_attributes` to make it clearer

essa:
@davidsiaw 
Thanks! 
That's much easier to read for me.

LGTM besides one more request.

Now, `prenullify_params` is clear about HOW but not about WHY this is needed.
Could you add a comment for it, such as "This is needed for existing services which has been created when the default of cpu is 512, not nil"?

davidsiaw:
Alright will do that

davidsiaw:
@essa I added a comment to the method so the next person will know why it was written

essa:
Thanks!
It will help future coming developers much!

k2nr:
This LGTM 👍 Is this ready for merge?

davidsiaw:
Yes it is

k2nr:
Thank you for fixing the issue!

PR #569, opened 2020-04-01
Make all one-off tasks set lang to UTF8

### LINKED ISSUE #568: LANG variable not set when cron runs
Right now barcelona sets the `LANG` variable if the `interactive` parameter to one-offs is set to true. However this is not true for cron jobs. It is extremely likely that tasks run as cron jobs are tested beforehand with `bcn run` and therefore with this variable set.

A bug https://github.com/degica/bizstation_api/pull/28 has recently come up because the developers wrote the program to rely on this behavior of `bcn run` and were surprised when a cron task did not have ENV vars read as utf-8.

https://github.com/degica/barcelona/blob/17cb387031d4895b27683dfe1b3cbd81520360ea/app/models/oneoff.rb#L33-L37

We should just set this variable all the time for one-offs, even when its a cron job, instead of just interactive all the time.

davidsiaw:
# Context

This PR makes it so `LANG=C.UTF-8` is set for all one-off tasks, not just `interactive: true`. I'm not actually sure if this is safe, I think it is, but please let me know if you are depending on LANG not being set.

Fixes #568

### DIFF
--- a/app/models/oneoff.rb
+++ b/app/models/oneoff.rb
@@ -34,7 +34,7 @@ class Oneoff < ActiveRecord::Base
             # Ideally Barcelona should not override LANG but because all official docker images
             # doesn't set LANG as UTF8 we can't use multi byte characters in
             # the interactive session without this override
-            environment: interactive ? [{name: "LANG", value: "C.UTF-8"}] : []
+            environment: [{name: "LANG", value: "C.UTF-8"}]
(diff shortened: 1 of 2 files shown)


degicat:
@essa please review this

davidsiaw:
I guess this should be good because #570 will be coming next

PR #570, opened 2020-04-01
Add params to add env vars to oneoff tasks

davidsiaw:
# Context

This PR adds a parameter to the one-off tasks endpoint to allow running one-off tasks with specific environment variables.

Fixes https://github.com/degica/barcelona-cli/issues/14

You need the new barcelona cli available here: https://github.com/degica/barcelona-cli/releases/tag/0.9.3

# How to test
<version of bcn cli>

1. Pull https://github.com/degica/barcelona-hello
2. Deploy the heritage
3. Try and run `bcn run -e hello -E FOO=bz bash`
4. Run `echo $FOO` and observe that it outputs `bz`

### DIFF
--- a/app/controllers/oneoffs_controller.rb
+++ b/app/controllers/oneoffs_controller.rb
@@ -9,7 +9,10 @@ class OneoffsController < ApplicationController
   def create
     interactive = !!params[:interactive]
     @oneoff = @heritage.oneoffs.create!(create_params)
-    @oneoff.run!(sync: !!params[:sync], interactive: interactive, started_by: "barcelona/#{current_user.name}")
+    @oneoff.run!(sync: !!params[:sync],
+                 interactive: interactive,
+                 started_by: "barcelona/#{current_user.name}",
+                 env_vars: env_var_params)
     json = if interactive
              certificate = @heritage.district.ca_sign_public_key(
                current_user,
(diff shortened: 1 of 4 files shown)


degicat:
@essa please review this

app/controllers/oneoffs_controller.rb
```diff
@@ -34,6 +37,23 @@ def create_params
...
+    params[:env_vars].each do |key, value|
+      if !key.is_a? String
+        raise ExceptionHandler::BadRequest.new("Keys in env_vars should be strings")
+      end
+      if !value.is_a? String
+        raise ExceptionHandler::BadRequest.new("Values in env_vars should be strings")
+      end
+    end
+
+    params[:env_vars].permit(params[:env_vars].keys)
```
essa, line 54: What does this line do?

davidsiaw: It turns the hash into a permitted params instance so it will be accepted by activerecord for mass assignment

essa: Thanks, I see. 
So it's not permitting, it's a conversion allowing every input from external.

I think it's better to use `permit!` as described in following article to warn that "we allow all the parameters from input here!".

https://www.aloucaslabs.com/miniposts/how-to-permit-all-parameters-in-rails-controller-strong-parameters

essa:
LGTM besides my comment.

Leave it if you think it's an idiom of Rails.

davidsiaw:
Thank goodness for tests. I legitimately broke something

PR #583, opened 2020-05-07
Make web_container_port configurable

### LINKED ISSUE #553: Portmapping stuck to pointing at container:3000

essa:
Resolves #553 with https://github.com/degica/barcelona-cli/pull/27.

I tested this with following barcelona.yml

```yaml
environments:
  production:
    name: hello
    image_name: docker.io/essa/barcelona-sample-hello
    services:
      - name: web
        service_type: web
        public: true
        cpu: 128
        memory: 256
        command: ruby hello.rb -p 8001 # Specify port for application to bind
        web_container_port: 8001 #  Specify port for connecting by reverse proxy
        listeners:
          - endpoint: essa-hello
            health_check_interval: 10
            health_check_path: /
            rule_conditions:
              - type: path-pattern
                value: '*'
```

`web_container_port` is used for `container_port` in port_mappings. So you can use it any application with fixed bind port.

### DIFF
--- a/spec/requests/create_heritage_spec.rb
+++ b/spec/requests/create_heritage_spec.rb
@@ -66,18 +66,27 @@ describe "POST /districts/:district/heritages", type: :request do
       expect(heritage["services"][0]["command"]).to eq "nginx"
       expect(heritage["services"][0]["force_ssl"]).to eq true
       expect(heritage["services"][0]["reverse_proxy_image"]).to eq "org/custom_revpro:v1.2"
-      expect(heritage["services"][0]["port_mappings"][0]["lb_port"]).to eq 3333
-      expect(heritage["services"][0]["port_mappings"][0]["container_port"]).to eq 3333
-      expect(heritage["services"][0]["port_mappings"][0]["host_port"]).to be_a Integer
-      expect(heritage["services"][0]["port_mappings"][0]["protocol"]).to eq "udp"
-      expect(heritage["services"][0]["port_mappings"][1]["lb_port"]).to eq 80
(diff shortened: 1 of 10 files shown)


spec/requests/create_heritage_spec.rb
```diff
@@ -66,18 +66,27 @@
       expect(heritage["services"][0]["command"]).to eq "nginx"
       expect(heritage["services"][0]["force_ssl"]).to eq true
       expect(heritage["services"][0]["reverse_proxy_image"]).to eq "org/custom_revpro:v1.2"
-      expect(heritage["services"][0]["port_mappings"][0]["lb_port"]).to eq 3333
```
essa, line 69: I need to update this because it depends on orders of `port_mappings`. But contents for test is the same as before.

essa:
@davidsiaw 
Can you review this?

spec/requests/create_heritage_spec.rb
davidsiaw: make sense. i had to deal with that too

PR #587, opened 2020-05-18
Install bcnd

davidsiaw:
It turns out the deployment PR was not complete because bcnd failed. https://github.com/degica/barcelona/actions/runs/107903526

This PR installs bcnd so it can be called


degicat:
@Resonious please review this

PR #594, opened 2020-06-18
Deploy on Build

davidsiaw:
The previous method of deploying didn't work because the commit of the master branch upon merch has a different SHA from when it was built in the branch. This commit just builds the commit itself again on master and then deploys it after that.


essa:
LGTM 👍 
Thanks!

essa:
@davidsiaw 
https://github.com/degica/barcelona/runs/782819911?check_suite_focus=true
It did invoke build step but not deploy.
Can you check it?

PR #597, opened 2020-06-29
Update dev env

k2nr:
While I was working on the firelens change in the last week's Open Hack Day, I realized that dev environment with docker-compose were broken so I fixed them.

This fix includes

- Setup VSCode dev environment
- Updated docker related files so docker-compose work
  - also deleted the spring container since it's not used
- Updated binstubs

This PR should not affect production environment.


k2nr:
@essa can you review? (asking you because I think you're using VSCode. sorry if you're no longer using it 😅 )

essa:
Sure, I'm using VSCode, will check it.

essa:
LGTM 👍 

I have tested it on my Mac and succeeded to ...

- Edit ruby file with solargraph in dev container
- Debug puma and rspec in dev container

スクリーンショット 2020-07-06 18 03 41


Only issue I have had is the spec file pattern in `launch.json`. I fixed it.

And I have to set the solargraph config in Setting -> Extentions -> Solargraph -> Command Path as `/usr/local/bundle/bin/solargraph`. This is the path of the solargraph binary in the container.

k2nr:
thanks!

PR #601, opened 2020-07-20
Add support for filtering Nginx logs

thoiberg:
Because there are GDPR requirements on sending Personally Identifiable Information (PII) to third party services, we want to make sure we are not logging any PII that gets sent as query params in requests. This PR uses the [Nginx JS Module](https://www.nginx.com/blog/introduction-nginscript/) to check query params in the incoming requests, and filter out any that match the Email regex.

I see this as a first step in supporting log filtering, if needed I think we can expand it and formalise an API around the JS modules, to make it easier to add new filters or instance specific filters. But I would rather wait until we have a clearer understanding of what we would need before trying to solve that problem 😄 . 

## How to Test
**Testing with filtering enabled**
1. Clone the branch
1. Run `docker-compose -f docker-compose.test.yml build`
1. Run `docker-compose -f docker-compose.test.yml up`
1. Using the browser (or curl) go to http://localhost:50132/?email=john.smith@smith.com
1. In the nginx logs you should see `email=[EMAIL FILTERED]`, like the screenshots

**Testing with filtering disabled**
1. Stop the docker containers session if it's still running
1. Change the [`FILTER_LOGS` param](https://github.com/degica/barcelona/pull/601/files#diff-71027439e4e13f5ebb931f5ea1f97b66R14) to "false" (or remove it entirely)
1. Run `docker-compose -f docker-compose.test.yml up`
1. Using the browser (or curl) go to http://localhost:50132/?email=john.smith@smith.com
1. In the nginx logs you should see `email=john.smith@smith.com`

## Screenshots
**with `FILTER_LOGS=true`**
Screen Shot 2020-07-20 at 3 08 52 pm

**with `FILTER_LOGS=false`**
Screen Shot 2020-07-20 at 3 09 05 pm


dockerfiles/reverse-proxy/README.md
```diff
@@ -0,0 +1,13 @@
+# reverse-proxy
+
+The reverse-proxy Dockerfile is the Barcelona supported reverse proxy (based on nginx) that is configured to cleanly integrate with the other Barcelona images.
```
thoiberg, line 3: Apologies if this is incorrect, I guessed a little 😄 . If there's a better description for the image let me know and I'll update the PR 😄

dockerfiles/reverse-proxy/templates/nginx.conf.erb
```diff
@@ -12,10 +14,12 @@ http {
...
   default_type  application/octet-stream;
 
-  log_format main '$proxy_protocol_addr - [$time_local] '
-                  '"$request" $status $body_bytes_sent '
-                  '"$http_referer" "$http_user_agent"';
-  access_log    /var/log/nginx/access.log main;
+  js_include maskQueryParams.js;
+  js_set $request_uri_masked maskQueryParams;
+
+  log_format main <%= log_format %>;
```
thoiberg, line 20: I'm not super happy with this way of setting the log format, because it messes up the formatting in the `nginx.conf` file:
image

I'm not sure how much of an issue that will be, since it's the generated file we probably won't be looking at too often, but if someones knows of a way to insert a multi line Ruby string via ERB and maintain formatting, let me know 😄 .

dockerfiles/reverse-proxy/docker-compose.test.yml
```diff
@@ -0,0 +1,17 @@
...
+        links:
+            - web
+        environment:
+            UPSTREAM_NAME: web
+            UPSTREAM_PORT: 80
+            FORCE_SSL: "false"
+            DISABLE_PROXY_PROTOCOL: "true"
+            FILTER_LOGS: "true"
+    web:
+        image: yeasy/simple-web
```
thoiberg, line 16: [yeasy/simple-web](https://hub.docker.com/r/yeasy/simple-web/) is a small web server that just keeps a log of requests that go to it. I was looking for the simplest web server I could find to avoid having to run any config, since we're really just trying to test how the nginx reverse proxy works, and not the website it's in front of.

For those interested, the webpage looks like this:
image

degicat:
@rramsden please review this

degicat:
@Resonious please review this

rramsden:
@thoiberg Thanks for tackling this one. I think before we review we need to discuss if this is something that should be filtered immediately or using something like a retention period (e.g. 14 days) where stored logs at a later period. The latter is much more complicated since we upload our logs to multiple services. Just let me check with CS and SRE teams if we can completely scrub this information.

rramsden:
@thoiberg Chatting with @k2nr our team has never looked at the nginx logs before. I think we can disable parameter logging completely in Nginx. Then for Rails and ELB we can work on a strategy for scrubbing those log sources.

rramsden:
@davidsiaw Any issues with disabling parameter logging with Nginx? We would still have logging with ELB and Rails I believe

rramsden:
@thoiberg Could you look at opening another PR and disable parameter logging in the template? Afterwards, assign @k2nr and @davidsiaw as reviewers. I'd like to close this one in favor of that PR if possible 🙏

davidsiaw:
I see no issue with disabling parameter logging.

thoiberg:
@rramsden sorry I don't think I fully understand. When you say "disable parameter logging in the template" do you mean moving the logic from the `render_template.rb` to the `nginx.conf.erb`, like:
```
log_format main '$proxy_protocol_addr - [$time_local] '
                '"$request" $status $body_bytes_sent '
                '"$http_referer" "$http_user_agent"';

log_format filtered '$proxy_protocol_addr - [$time_local] '
                    '"$request_method $request_uri_masked $server_protocol" '
                    '$status $body_bytes_sent '
                    '"$http_referer" "$http_user_agent"'

access_log /var/log/nginx/access.log  <%= ENV['filter_logs]' == 'true' ? 'filtered' : 'main' %>;
```

rramsden:
@thoiberg Yep, that's what I am thinking. We would want to replace `$request` variable here which evaluates to something like `GET /path?name=hogehoge&arg2=fugafuga HTTP/1.1` with something like `$request_method $uri $server_protocol` which would write to logs without query parameters.

@davidsiaw Does that sound right to you? I think that's what we want 🤔

rramsden:
Closing this one for now 🙇

PR #602, opened 2020-07-20
Do not push images for PRs

davidsiaw:
# Context

Our docker builds require a secret to push to docker right now. So this PR changes it to only check that it builds and does not attempt to login and push.


degicat:
@rramsden please review this

PR #603, opened 2020-07-22
Allow underscores to be passed in as review app subject

jdgc:
This is kind of a selfish PR, but I tend to use snake_case naming scheme for git branch names which does not pass the review app subject validation.

Underscores aren't legal characters in subdomains, so in this PR we check the subject for them and convert to hyphens if necessary.

Feel free to reject if handling underscores is deemed unnecessary

### DIFF
--- a/app/models/review_app.rb
+++ b/app/models/review_app.rb
@@ -4,7 +4,7 @@ class ReviewApp < ApplicationRecord
 
   attr_accessor :image_name, :image_tag, :services, :before_deploy, :environment
   validates :subject, :image_name, :image_tag, :retention, :services, presence: true
-  validates :subject, format: {with: /\A[a-z0-9][a-z0-9-]*[a-z0-9]\z/}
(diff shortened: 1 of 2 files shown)


app/models/review_app.rb
```diff
@@ -68,4 +68,9 @@ def computed_environment
       {name: "BARCELONA_REVIEWAPP_DOMAIN", value: domain}
     ]
   end
+
+  # replace underscores if present - not valid as subdomain
+  def sanitized_subject
+    subject.gsub('_', '-')
```
davidsiaw, line 74: If we are going to do this lets just remove the validation and go ham `gsub(/[^a-z0-9]+/, '-')`

What do you think?

jdgc: this would let nonsense through to be interpreted as hyphens (eg. #!$%). If that's acceptable I think its okay

degicat:
@davidsiaw please review this

degicat:
@Yuuki77 please review this

davidsiaw (approved):
Seems good. lets see how this goes.

Yuuki77:
LGTM!

PR #605, opened 2020-08-04
Update to use vault-degica

davidsiaw:
This PR makes a bunch of changes and fixes to update barcelona to start working with vault again.

Also includes some refactoring and test cleanups.

## How to test
1. Set up a vault instance. Note down the host and port. (Internal reviewers can use the vault-degica repo to do this)
2. Create a github authorization method on vault, then create an organization or use an existing one. 
3. Using that host and port start up barcelona using:
```
# assuming your vault instance is located locally and on port 8200
VAULT_HOST=http://localhost VAULT_PORT=8200 ORGANIZATION=ORGNAME kaiser up
```
In the organization, make sure a team called `testers` exists and contains a user account that you have access to.
4. Create a personal token with that user account with **only** the permission `read:org`
5. ~Download https://github.com/degica/barcelona-cli/releases/tag/rc-0.9.7~ Get [this](https://github.com/degica/barcelona-cli/pull/38) version of barcelona-cli
6. Login using Vault with your personal token
```
./bcn login --auth vault --vault-token <your token> --vault-url https://vault.localhost.labs.degica.com https://barcelona.localhost.labs.degica.com
```
7. Observe that the command fails with `You are not authorized to do that action`
8. Set up the user and a policy by going:
```
vault auth enable github
vault write auth/github/config organization=ORGNAME
vault write auth/github/map/users/USERACCOUNT value=testers

/ $ cat <<HCL >> /tmp/testers.hcl
> path "secret/Barcelona/ORGNAME/*" {
>   capabilities = ["create", "update", "read", "delete", "list"]
> }
> HCL
cat <<HCL >> /tmp/testers.hcl
```
9. Attempt to login again with
```
./bcn login -a vault http://barcelona.lvh.me
```
10. Observe that the command succeeds and creates a `.bcn` directory in your $HOME
11. Observe that you can also proceed to perform different operations such as 
```
./bcn district list
```


degicat:
@k2nr please review this

app/auth/vault_auth.rb
```diff
@@ -1,74 +1,63 @@
...
+
   def self.enabled?
-    ENV['VAULT_URL'].present?
+    vault_url.present?
   end
 
   def login
-    authenticate
+    req = Net::HTTP::Post.new("/v1/auth/github/login")
+    req.body = { token: vault_token }.to_json
```
k2nr, line 12: So this is not a vault token but GitHub's personal token, right? I think getting a vault token should happen on client side so
1. Client calls Vault's `/v1/auth/github/login` with GH personal access token to get a vault token
2. Client pass the vault token to Barcelona
3. Barcelona calls Vault APIs with the passed vault token

davidsiaw: I see. So bcn cli grabs the token and passes it to barcelona. That makes sense. Will change

k2nr: yeah, so that's the background of this PR https://github.com/degica/barcelona/pull/615 with this meta API, cli can know what to do and where to request

davidsiaw:
@k2nr I have updated this if you want to have another look at it. Also you will need the modified client https://github.com/degica/barcelona-cli/pull/38 to use it this way

app/auth/vault_auth.rb
```diff
@@ -1,74 +1,49 @@
...
 
   def login
     authenticate
   end
 
   def authenticate
-    res = lookup
-    username = res.dig("data", "meta", "username")
-    @current_user = User.find_or_create_by!(name: username) if username
+    user = User.new(name: "vault-user-#{vault_token.hash.to_s(16)}")
```
k2nr, line 15: So when vault auth is used, `User` is not stored in DB, correct? I think that is fine because all authentication and authorization are done by Vault basically and Barcelona does not need to store users.
One concern here is, when we want to audit (who did what), there's no indication to identify "who". So I suggest to include `data.meta.username` field in the user name. 
What do you think?

k2nr:
Sorry for the late review.
Just one comment, everything else LGTM 👍 Could you take a look.

app/auth/vault_auth.rb
davidsiaw: That sounds like a very good idea.

davidsiaw: @k2nr I revisited this, but it turns out the `data.meta.username` is in the reply from authenticating with vault (i.e. the reply to sending the github token to vault). Since we are doing this from the CLI now, we do not send this data to vault (we only provide barcelona with the client token).

I tried looking for it but vault does not seem to have an API to interrogate the user name from a client_token. This seems to make sense because there may be some auth methods where there are no user names (token for example)

davidsiaw:
Was just testing barcelona on aws and have one more issue to fix with the users. We need to save users after all because we need to save the public key of users.

k2nr:
ah I see, you're right

davidsiaw:
Please see #632 for additions that load the username from vault.

davidsiaw:
@k2nr I will merge this in, and subsequent PRs will be able to address the following:

- saving users
- using github and vault at the same time
- loading the username from vault

k2nr:
I see. LGTM 👍 please go ahead and merge

PR #614, opened 2020-09-04
Vault fixes

k2nr:
This PR contains several fixes related to vault backend.

## Set `use_ssl` if the vault URL is https 

https://github.com/degica/barcelona/commit/52e398f9e168c1e561b16edc2b050259f52132cd

it was not set so https url couldn't be used

## Add non-shallow routes

https://github.com/degica/barcelona/pull/614/commits/ba64ac4f3856652d6d531f76dd5462a2893140d8

Currently heritage routes are shallow so district name is not in a path e.g. `/v1/heritages/heritage-name`.
This makes it harder to set vault permissions because it is a typical use case to allow teams to do all operations to a specific district.

I added new non-shallow routes. I didn't remove the shallow routes because it breaks CLI. Once this is merged, we need to update CLI to use the non-shallow routes and then we can remove the shallow routes.

## Improved auth backend selection

https://github.com/degica/barcelona/pull/614/commits/3b124d3cb30a8bc79b78db725dc5649d1c7edeae

Instead of relying on `X-Vault-Token` header, now backend is enabled / disabled by the following environment variables

- if `VAULT_URL` is set, vault auth is enabled
- if `GITHUB_ORGANIZATION` is set, github auth is enabled
- if both are set, vault is chosen

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,10 +27,13 @@ class ApplicationController < ActionController::API
   end
 
   def auth_backend
-    @_auth_backend ||= if request.headers['HTTP_X_VAULT_TOKEN']
-                         VaultAuth.new(request)
-                       else
-                         GithubAuth.new(request)
-                       end
+    @_auth_backend ||= begin
+      backend_class = if VaultAuth.enabled?
+                        VaultAuth
+                      elsif GithubAuth.enabled?
+                        GithubAuth
+                      end
+      backend_class.new(request)
+    end
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,6 +21,18 @@ Rails.application.routes.draw do
         post "/releases/:version/rollback", to: "releases#rollback"
       end
 
+      resources :heritages, except: [:new, :edit] do
+        post   :env_vars, on: :member, to: "heritages#set_env_vars"
+        delete :env_vars, on: :member, to: "heritages#delete_env_vars"
+
+        post "/trigger/:token", to: "heritages#trigger"
+        resources :oneoffs, only: [:show, :create]
+
+        get "/releases", to: "releases#index"
+        get "/releases/:version", to: "releases#show"
+        post "/releases/:version/rollback", to: "releases#rollback"
+      end
+
       resources :endpoints, except: [:new, :edit]
       resources :notifications, except: [:new, :edit]
     end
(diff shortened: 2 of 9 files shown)


degicat:
@davidsiaw please review this

spec/rails_helper.rb
```diff
@@ -54,8 +54,6 @@
     Aws.config[:stub_responses] = true
 
     ENV['GITHUB_ORGANIZATION'] = 'degica'
-    ENV['VAULT_URL'] ||= "http://my-vault.com"
```
k2nr, line 57: removed this because if this env is set the vault auth is used for all test cases. we want to use GitHub auth except for test cases that tests vault auth

config/routes.rb
```diff
@@ -21,6 +21,18 @@
...
+        post   :env_vars, on: :member, to: "heritages#set_env_vars"
+        delete :env_vars, on: :member, to: "heritages#delete_env_vars"
+
+        post "/trigger/:token", to: "heritages#trigger"
+        resources :oneoffs, only: [:show, :create]
+
+        get "/releases", to: "releases#index"
+        get "/releases/:version", to: "releases#show"
+        post "/releases/:version/rollback", to: "releases#rollback"
+      end
```
davidsiaw, line 34: If you expand the top you'll find nearly exactly the same code. Is this a duplicate?

app/controllers/application_controller.rb
```diff
@@ -27,10 +27,13 @@ def append_info_to_payload(payload)
...
-                         GithubAuth.new(request)
-                       end
+    @_auth_backend ||= begin
+      backend_class = if VaultAuth.enabled?
+                        VaultAuth
+                      elsif GithubAuth.enabled?
+                        GithubAuth
+                      end
+      backend_class.new(request)
+    end
```
davidsiaw, line 37: Can we test this just for safeness?

davidsiaw (changes requested):
Looks great except for a few small things

config/routes.rb
k2nr: you mean I can do like below?

```ruby
[true, false].each do |shallow|
  resources :heritages, shallow: shallow, except: [:new, :edit] do
    # ...
  end
end

app/controllers/application_controller.rb
k2nr: added

davidsiaw (approved):
LGTM

PR #621, opened 2020-10-06
Add ability to set CPU and Memory for scheduled task

davidsiaw:
Previously it was not possible to set a scheduled task's Memory or CPU usage, and it was pinned at 512MB of RAM and 128 units of CPU. This needs to change as some tasks require more memory.

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -76,6 +76,8 @@ class HeritagesController < ApplicationController
       :image_name,
       :image_tag,
       :before_deploy,
+      :scheduled_task_cpu,
+      :scheduled_task_memory,
       services: [
         :name,
         :cpu,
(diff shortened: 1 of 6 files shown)


degicat:
@Iorin0225 please review this

davidsiaw:
@Iorin0225 sorry I meant to put @k2nr on this review but I think I messed up the command. I will remove you as reviewer for this PR

k2nr:
I think it's better to set cpu and memory per task like this. Also, same as services, can we change cpu default to `nil`?
```
scheduled_tasks:
  - schedule: rate(5 minutes)
    command: some_task
    memory: 1024
  - ...

davidsiaw:
> I think it's better to set cpu and memory per task like this

This is a little more complicated to do because we store this information per heritage. This will involve changing the model and moving a bunch of things around. For now the idea is just to make it modifiable.

> can we change cpu default to nil? 

We can, I just kept the CPU the same so we don't change too many things at once. I can put up a PR that sets the scheduled task CPU to default to `nil`. I can't think of any possible problems with that, but there could be problems that would emerge that I don't know.

PR #623, opened 2020-10-09
Fix bootstrap script

davidsiaw:
A small change to Rails 5.2 meant that running rake tasks now checks for the presence of `SECRET_KEY_BASE` all the time. This probably wasn't the case when the script was first written.

### DIFF
--- a/lib/tasks/bootstrap.rake
+++ b/lib/tasks/bootstrap.rake
@@ -57,6 +57,8 @@ namespace :bcn do
       ENV["BOOTSTRAP_DATABASE_URL"] = "postgresql://#{stack.db_user}:#{db_password}@#{db_endpoint}:5432/#{stack.db_name}"
     end
 
+    secret_key_base = SecureRandom.hex(64)
+
     # Run oneoff task that runs inside the VPC and creates barcelona service and endpoint
     heritage = district.heritages.new(
       name: "barcelona-bootstrap",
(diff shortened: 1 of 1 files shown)


degicat:
@Resonious please review this

lib/tasks/bootstrap.rake
```diff
@@ -142,7 +145,7 @@ EOS
           {key: "RAILS_LOG_TO_STDOUT", value: "true", secret: false},
           {key: "GITHUB_ORGANIZATION", value: ENV['GITHUB_ORGANIZATION'], secret: false},
           {key: "DATABASE_URL",  value: ENV["DATABASE_URL"], secret: true},
-          {key: "SECRET_KEY_BASE", value: SecureRandom.hex(64), secret: true},
+          {key: "SECRET_KEY_BASE", value: secret_key_base, secret: true},
```
Resonious, line 148: I may be crazy, but it looks to me like this `secret_key_base` local variable is defined in the task `bcn:bootstrap`, and right here it is being used by the task `bcn:bootstrap:remote`. Those are different Ruby scopes so I suspect this line here will error out.

We can probably just keep the inline call to `SecureRandom.hex(64)` here, right?

Resonious (changes requested):
@davidsiaw Change makes sense but I think there's a small error. Let me know if I'm just seeing things wrong :laughing:

lib/tasks/bootstrap.rake
davidsiaw: Oh yeah. That is odd. I wonder why this worked... It may have been because the variable was re-evaluated. You're right. It can just be the same key base here too.

PR #629, opened 2020-11-05
Save users

davidsiaw:
This PR follows #605 to fix saving public keys required for running `bcn run`.

This PR also adds some tests that were missing for users.

This PR will be ready review after #605 is merged.

## How to test
1. Deploy branch `vault-dev` and login through vault
2. Try to run `bcn run -e test echo hello`
3. Note that it fails
4. Deploy this branch
5. Run `bcn run -e test echo hello`
6. Note that it still fails
7. Login again with vault
8. Run `bcn run -e test echo hello`
9. Note that it passes

PR #630, opened 2020-11-05
Allow dual use

davidsiaw:
Allows use of both Vault and Github backends


k2nr:
merged #629 can you rebase?

PR #633, opened 2020-11-13
Control container count

### LINKED ISSUE #565: API to increase/decrease instance/container count
We need an API to increase and decrease instance or container counts in barcelona.

davidsiaw:
This PR introduces the ability to set container count using barcelona

Fixes #565

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -68,6 +68,15 @@ class HeritagesController < ApplicationController
     render json: @heritage
   end
 
+  def update_service_scale
+    service.desired_container_count = desired_container_count
+    service.save!
+    @heritage.save_and_deploy!(without_before_deploy: true,
+                               description: "Change service scale #{service.name} to #{desired_container_count}")
+
+    render json: @heritage
+  end
+
   def permitted_params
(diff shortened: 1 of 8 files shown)


degicat:
@essa please review this

essa:
@davidsiaw 
Great!

I have tested it local and succeeded to increase task count by following command.

```
$ bcn api post /districts/essa-test/heritages/hello/services/web/count '{"desired_container_count": 3}' 
```

But I have an issue.

Before the command, task count is 1. So I expected this commands start two more tasks.
But it started three tasks and stopped the old task after new tasks started.

So, if we are running 80 tasks and increase it to 120, we need enough container instances for 200 (80+120) tasks to complete the update.

I think it's not critical, only we need more caution on available resources when we use this api.

@k2nr 
Do you know any workarounds for it?

davidsiaw:
@essa Ah yes. I think I know why, because it creates a new task definition when it changes the task count. I can try and solve this situation in another PR

essa:
@davidsiaw 

>I can try and solve this situation in another PR

Great!

Then LGTM for this PR.

essa (approved):

PR #638, opened 2020-11-26
Enable honeybadger reporting if API key is present

rramsden:
## Context

This pull requests allows `HONEYBADGER_API_KEY` to enable error reporting with Honeybadger.
This is a nice service to see errors in real-time so they can be debugged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'gibberish'
 gem 'octokit'
 gem 'delayed_job_active_record'
 gem 'pundit'
+gem 'honeybadger', require: false
 gem 'jbuilder'
 gem 'lograge'
 gem 'net-ssh'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -127,6 +127,7 @@ GEM
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     hashdiff (0.3.8)
+    honeybadger (4.7.2)
     i18n (1.8.2)
       concurrent-ruby (~> 1.0)
     jaro_winkler (1.5.4)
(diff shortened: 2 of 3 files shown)


rramsden:
@k2nr @davidsiaw What do you think? I'm not sure of a nicer way to get error reporting in Barcelona as its open source and we don't want too much dependency bloat happening. This PR would only enable it if the environment variable is present.

davidsiaw:
@k2nr I thought about it a little and I don't think its a big issue, in particular instead of having to dig through the logs this acts as an implicit filter to only display errors so that is a benefit.

It is probably not ideal to have barcelona tied to honeybadger but since this is an optional dependency I don't think it will greatly hurt barcelona. In the future we can have a more generic error reporting system for barcelona but for now this will probably do.

Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'gibberish'
 gem 'octokit'
 gem 'delayed_job_active_record'
 gem 'pundit'
+gem 'honeybadger', require: false
```
davidsiaw, line 21: I suspect honeybadger has to be placed at the bottom of the Gemfile for [reasons](https://github.com/degica/hats/blob/423c057916fa9d814780503e276b741922e93a99/Gemfile#L356-L357).

davidsiaw (changes requested):
Looks good to me. Maybe just one thing

k2nr:
Looks good. I don't think this is an issue as long as honeybadger is optional.

Gemfile
rramsden: @davidsiaw I think this is fine since we are not including the gem yet `require: false`. The gem gets loaded when Rails initializer is booted... maybe this is okay?

davidsiaw: Great. That is why. Thanks.

PR #639, opened 2020-11-26
Use S3 for Stack templates

davidsiaw:
Use S3 for stack templates to overcome 51k size limit.

## How to test
Need to find a way to make the template so big it usually wouldn't fit in 51k.


degicat:
@essa please review this

davidsiaw:
Hi, Nakajima-san, this requires a person who can use S3 and cloudformation to check that it is working. I have tested it myself already, let me know if you want access to our test barcelona on slack to try this out if you want.

essa (approved):
LGTM 👍

PR #643, opened 2021-01-06
Exclude variables that exist in new environment

k2nr:
Consider the following case:

### Case 1

- non-secret env var `VARIABLE` exists as a legacy `EnvVar`
- secret environment `VARIABLE` exists in `environment`

### Case 2

- secret env_var `VARIABLE` exists as a legacy `EnvVar`
- non-secret environment `VARIABLE` exists in environment
- 
With these cases Barcelona adds `VARIABLE` to both `Environment` and `Secrets` of task definition.

This PR solves the above issue by removing legacy environment variables that also exist in `environment`

### DIFF
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -317,14 +317,16 @@ class Heritage < ActiveRecord::Base
   end
 
   def environment_set
-    legacy_env_vars = env_vars.where(secret: false).map { |e| [e.key, e.value] }.to_h
+    legacy_env_vars = env_vars.where(secret: false).
+                               where.not(key: environments.pluck(:name)).
+                               map { |e| [e.key, e.value] }.to_h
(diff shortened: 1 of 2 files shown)


davidsiaw (changes requested):
Can we write a test that fails with the old code?

k2nr:
sure

k2nr:
Done. confirmed that the spec fails without the fix

k2nr:
OK this is ready for review

degicat:
@davidsiaw please review this

davidsiaw (changes requested):
Just a few small changes for readability

k2nr:
Thanks for the suggestions! I picked your suggestions.

davidsiaw (approved):
LGTM

PR #645, opened 2021-01-15
Set correct value/value_from on update

k2nr:
## Problem

Currently converting raw value environment to secret doesn't work.

Let's say we have the following raw environment:

```
environment:
  - name: SOME_ENV
    value: "some value"
```

and we want to change the environment to secret:

```
environment:
  - name: SOME_ENV
    ssm_path: "path/to/ssm"
```

This update is ignored. the raw "some value" is kept using. The opposite case (`ssm_path` -> `value`) also doesn't work.

This PR fixes the above issue.


degicat:
@davidsiaw please review this

davidsiaw (approved):
Tested it and it works. LGTM

PR #646, opened 2021-01-15
Fix user validation over multiple auths

davidsiaw:
The PR addresses an issue that came up when testing vault, the uniqueness condition was changed in #408 : https://github.com/degica/barcelona/pull/408/files#diff-ba6173ca23a9f06e9d61c8ce8b2edf1d7f30281eaa786c6f5d416f97cffd1760 

But the model was not modified to reflect this. This PR does that.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,7 +5,7 @@ class User < ActiveRecord::Base
 
   serialize :roles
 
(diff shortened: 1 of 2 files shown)


degicat:
@FTLam11 please review this

FTLam11 (approved):
L:zany_face:TM, thanks for this PR!

PR #648, opened 2021-01-21
Use non-shallow path for vault auth

k2nr:
For heritages paths, bcn cli uses shallow paths e.g. `"/v1/heritages/name"` instead of `"/v1/districts/district/heritages/name"` because when `bcn run -H heritage-name rails c` is executed, the cli does not know what district the heritage belongs to.
But this shallow path doesn't work well with vault auth because often we want to allow developers full access to a certain district.
For example, if we want to allow developers to access all heritages in a staging district, we want to declare below capability
```
  path "secret/Barcelona/degica/v1/districts/staging*" {
    capabilities = ["create", "update", "read", "delete", "list"]
  }
```

This PR changes vault auth to use non-shallow paths for heritage paths

### DIFF
--- a/app/auth/vault_auth.rb
+++ b/app/auth/vault_auth.rb
@@ -33,7 +33,7 @@ class VaultAuth < Auth
   end
 
   def authorize_action
-    if !cap_probe.authorized?(request.path, request.method)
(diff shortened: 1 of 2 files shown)


degicat:
@davidsiaw please review this

davidsiaw (approved):
LGTM

PR #651, opened 2021-02-02
Fix barcelona deploys failing sometimes

### LINKED ISSUE #650: Fix occasional deploy failures
Since #639 we have started loading up templates for deployments via S3. However, S3 is not immediately consistent, and so there can be a time after uploading when the file is not available for read. This manifests itself as an error like [this one](https://app.datadoghq.com/logs?cols=core_host%2Ccore_service&from_ts=1611882600000&index=&live=false&messageDisplay=inline&query=service%3Abarcelona2-worker&stream_sort=desc&to_ts=1611882900000&event=AQAAAXdLtNgY-MERAgAAAABBWGRMdE5wU0FBQlBRVzMtVkZaVXJRQUE). This phenomenon is [well known](https://stackoverflow.com/questions/8856316/amazon-s3-how-to-deal-with-the-delay-from-upload-to-object-availability). We can probably add a delay or retry to ensure that deploys don't just die for this reason.

davidsiaw:
~This PR introduces a wait after put object as a way to ensure that S3 has updated and made the object we just placed available.~

Turns out it was a method that should have been memoized but was not.

Fixes #650 

# How to test
This is one of those bugs that happens only occasionally due S3's behavior so this is difficult to reproduce. The normal path however, should not be any different.

1. Deploy an application
2. Check that it successfully deployed

### DIFF
--- a/lib/cloud_formation/executor.rb
+++ b/lib/cloud_formation/executor.rb
@@ -56,13 +56,29 @@ module CloudFormation
     end
 
     def upload_to_s3!
-      resp = @s3_client.put_object({
-        body: stack.target!,
+      params = {
         bucket: @bucket,
         key: template_name,
+      }
+
+      resp = @s3_client.put_object({
+        body: stack.target!,
+        **params
       })
-      Rails.logger.info "Uploaded stack template to bucket"
       Rails.logger.info resp
+
+      begin
+        @s3_client.wait_until(:object_exists, params, 
+          before_wait: -> (attempts, response) do
+            Rails.logger.info "Waiting for stack template to be uploaded"
+          end
+        )
+      rescue Aws::Waiters::Errors::WaiterFailed => e
+        Rails.logger.warn "Upload failed: #{e.message}"
+        raise e
+      end
+
+      Rails.logger.info "Uploaded stack template to bucket"
     end
 
     def stack_options
--- a/lib/cloud_formation/executor.rb
+++ b/lib/cloud_formation/executor.rb
@@ -67,6 +67,7 @@ module CloudFormation
       })
       Rails.logger.info resp
 
+      sleep(5)
       begin
         @s3_client.wait_until(:object_exists, params, 
           before_wait: -> (attempts, response) do
--- a/lib/cloud_formation/executor.rb
+++ b/lib/cloud_formation/executor.rb
@@ -8,6 +8,8 @@ module CloudFormation
       @client = district.aws.cloudformation
       @s3_client = district.aws.s3
       @bucket = district.s3_bucket_name
+
+      upload_to_s3!
     end
 
     def describe
@@ -68,7 +70,7 @@ module CloudFormation
       Rails.logger.info resp
 
       Rails.logger.info "Waiting for stack template to be uploaded"
-      sleep(5)
+      sleep(30)
       begin
         @s3_client.wait_until(:object_exists, params, 
           before_wait: -> (attempts, response) do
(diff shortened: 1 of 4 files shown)


degicat:
@essa please review this

spec/lib/cloud_formation/executor_spec.rb
```diff
@@ -11,16 +11,29 @@
...
 
     it "updates the stack directly if false" do
       expect(s3).to receive(:put_object)
+      expect(s3).to receive(:wait_until).with(:object_exists, anything, anything)
       expect(client).to_not receive(:create_change_set)
       expect(client).to receive(:update_stack)
       executor.update(change_set: false)
     end
+
+    it "creates a change set if true" do
```
essa, line 28: You forgot to change the comment?

essa (approved):
LGTM 👍 besides my comment

spec/lib/cloud_formation/executor_spec.rb
davidsiaw: oops. Will fix this.

davidsiaw:
@essa Sorry I'm making you review this twice.

I did some deeper testing to make sure I really fixed the issue, I made it so my stack was deployed in other regions but activate CF here so I can make sure the delay is very long, but it turns out it didn't matter where I deployed it!

I introduced a delay to see if that compensated and it caused it to fail even more often. I was so puzzled yesterday I was dreaming about the problem.

It turns out I have a function to provide `template_name` but it changes every second! This was why it works sometimes and fails sometimes. If the processing was not fast enough (within the second) then the template name will change, and CF will be given the wrong name, and thus S3 will say Access Denied because it is not the owner of the bucket.

Was a very stupid bug.

essa:
Sorry, I missed it too.
I should have noticed on the first review.

lib/cloud_formation/executor.rb
```diff
@@ -52,25 +52,42 @@ def create_change_set(type: "UPDATE")
...
 
     def stack_options
       upload_to_s3!
       {
         stack_name: stack.name,
         capabilities: ["CAPABILITY_IAM"],
-        template_url: "https://#{@bucket}.s3.amazonaws.com/#{template_name}"
+        template_url: "https://s3.#{stack.region}.amazonaws.com/#{@bucket}/#{template_name}"
       }
     end
```
essa, line 92: @davidsiaw 
How about making `template_name` local variable instead of an instance_method?
```ruby
def stack_options
  template_name = "stack_templates/#{stack.name}/#{Time.current.strftime("%Y-%m-%d-%H%M%S")}.template"
  upload_to_s3!(template_name)
  {
        stack_name: stack.name,
        capabilities: ["CAPABILITY_IAM"],
        template_url: "https://s3.#{stack.region}.amazonaws.com/#{@bucket}/#{template_name}"
  }
end
```

I think this makes more clear about the role or scope of the `template_name`.

- It will be used only for communication between uploading and executing
- We don't have to expose it to other classes or save it for future use

davidsiaw: Its used in two places. if I do that the code will be duplicated.

davidsiaw: @essa Its used in two places. if I do that the code will be duplicated.

essa: @davidsiaw 
I meant removing the method and making it a local variable for passing as a parameter.
Where is the duplication?

davidsiaw: @essa my bad, I misread your code. That is a good idea. will do that.

PR #652, opened 2021-02-03
Fix internal server error

davidsiaw:
## Context

For a while now when you attempted to deploy a currently deploying stack, Barcelona would inexplicably return `internal_server_error`. After investigating the logs it turned out that a change was made in the past to remove the `heritage` parameter on `DeployRunnerJob#notify`, but one line of code was not updated. This PR fixes that and tests that code branch.

### DIFF
--- a/app/jobs/deploy_runner_job.rb
+++ b/app/jobs/deploy_runner_job.rb
@@ -9,7 +9,7 @@ class DeployRunnerJob < ActiveJob::Base
     @heritage = heritage
     heritage.with_lock do
       if other_deploy_in_progress?(heritage)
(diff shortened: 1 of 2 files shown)


degicat:
@AlessandroMinali please review this

AlessandroMinali (approved):
Look good to me.

What does heritage mean in this context though? Naming feels odd to me.

davidsiaw:
@AlessandroMinali Heritage just means an application (task definitions for containers for scheduled jobs, web and job workers). A heritage sits on a district (a VPC, bastion server, ec2 instances etc...). I agree that its a little odd but its just a name

Issue #616, opened 2020-09-09
Endpoint names are global


davidsiaw:
If you create an endpoint named A in district 1, you cant make an endpoint named A in district 2, even though there is no reason you shouldn't be able to.

## Possible fix

We should use the district parameter when creating an endpoint, and require that all operations on the endpoint require the district to be specified.

davidsiaw:
We will revert #660 because it introduces an unexpected change to the way barcelona behaves.

If you go

```
bcn endpoint create hello -d staging
```

Barcelona then creates an endpoint called `staging-hello`.

If in the past another district `test` has and endpoint called `staging-hello`, this PR causes barcelona to either fail or worse overwrite the endpoint name (we don't know what will happen)

This PR did not resolve the global endpoint name issue and caused the above new issue, so we are reverting it.

PR #658, opened 2021-03-08
fix that endpoint names are global

Yuuki77:
Fixes https://github.com/degica/barcelona/issues/616

If you create an endpoint named A in district 1, you cant make an endpoint named A in district 2, even though there is no reason you shouldn't be able to.

How to test
I think looking at the test should be ok

### DIFF
--- a/spec/models/endpoint_spec.rb
+++ b/spec/models/endpoint_spec.rb
@@ -15,10 +15,20 @@ describe Endpoint do
       endpoint.save!
     end
 
-    it "deltes cloudformation stack" do
+    it "deletes cloudformation stack" do
       endpoint.save!
       expect_any_instance_of(CloudFormation::Executor).to receive(:delete)
       endpoint.destroy!
+    end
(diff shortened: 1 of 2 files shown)


spec/models/endpoint_spec.rb
```diff
@@ -15,13 +15,23 @@
...
 
-    it "deltes cloudformation stack" do
+    it "deletes cloudformation stack" do
       endpoint.save!
       expect_any_instance_of(CloudFormation::Executor).to receive(:delete)
       endpoint.destroy!
     end
   end
 
+  describe "#validations" do
```
Yuuki77, line 25: `shoulda-matchers` doesn't support scope with association.
so I wrote the code like this.
https://github.com/thoughtbot/shoulda-matchers/issues/814

Yuuki77:
actually this is not going to work because  endpoint names is used for load balancer name and name should be unique in region.

Issue #617, opened 2020-09-10
You can mis-deploy a heritage into the wrong district


davidsiaw:
If you have district A and district B, it is possible to go

```
bcn create -e prod -d A
```

and then go

```
bcn create -e prod -d B
```

This is due to the fact that names are global but they can only be in one district. Deleting the heritage in B leaves the heritage in A still running but in barcelona its considered deleted. You can salvage this by creating the heritage in A again.

This does not cause trouble however until you use secret keys. If you create the heritage A and then set a secret key, creating the heritage in B will make the container fail to deploy. It will also cause you to not be able to run `bcn run` and instead get this error: `AccessDenied: Access Denied`

This problem is related to https://github.com/degica/barcelona/issues/616. We need to create a new API that requires both heritage and district names and deprecate the old one. This also means we need to modify the Barcelona CLI to provide both names.

## Possible fix
For now we need to prevent heritages of the same name created from being created.

The next thing we need to do is make barcelona handle heritages keyed by district. This means barcelona needs to complain if it is asked to perform operations on a heritage name which is shared by two heritages in different districts.

davidsiaw:
Reopening because merged PR does not fix the problem.

PR #659, opened 2021-03-09
throw error if heritage name already exits

Yuuki77:

Currently you can do

If you have district A and district B, it is possible to go

```
bcn create -e prod -d A
```

and then go

```
bcn create -e prod -d B
```

For now we need to prevent heritages of the same name created from being created.

### Note
Originally, I tried to do this inside BuildHeritage class, but the BuildHeritage class is also used in the reviewApp, and is used in both the create and find. So, I decided to handle it in the controller.

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -12,6 +12,10 @@ class HeritagesController < ApplicationController
   end
 
   def create
+    if Heritage.find_by(name: params[:name]).present?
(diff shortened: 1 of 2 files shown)


degicat:
@davidsiaw please review this

davidsiaw:
Hmm. I think people should be allowed to create another heritage with the same heritage name in a different district. Is there a way to accommodate this?

davidsiaw (approved):
Thanks. This is a good temporary solution but I think we should seek a permanent solution in the meantime!

PR #660, opened 2021-03-15
Fix that endpoint names are global

### LINKED ISSUE #616: Endpoint names are global

Yuuki77:
Fixes #616

If you create an endpoint named A in district 1, you cant make an endpoint named A in district 2, even though there is no reason you shouldn't be able to.

When creating an endpoint, we now add the district name as a prefix.

When `load_endpoint`, for compatibility with existing endpoints, I changed it to first look for the endpoint by name only, and if not found, look for it by heritage and name.

### DIFF
--- a/app/controllers/endpoints_controller.rb
+++ b/app/controllers/endpoints_controller.rb
@@ -8,12 +8,12 @@ class EndpointsController < ApplicationController
   end
 
   def create
-    endpoint = @district.endpoints.create!(permitted_params)
+    endpoint = @district.endpoints.create!(create_params)
     render json: endpoint
   end
 
   def update
-    @endpoint.update!(permitted_params)
+    @endpoint.update!(update_params)
     render json: @endpoint
   end
 
@@ -28,7 +28,9 @@ class EndpointsController < ApplicationController
 
   private
 
-  def permitted_params
+  def create_params
+    params[:name] =  name_prefix + params[:name]
+
     params.permit(
       :name,
       :public,
@@ -37,11 +39,26 @@ class EndpointsController < ApplicationController
     )
   end
 
+  def update_params
+    params.permit(
+      :certificate_id,
+      :ssl_policy
+    )
+  end
+
   def load_endpoint
-    @endpoint = @district.endpoints.find_by!(name: params[:id])
+    @endpoint = @district.endpoints.find_by(name: params[:id])
+    unless @endpoint.present?
+      @endpoint = @district.endpoints.find_by!(name: name_prefix + params[:id])
+    end
+    @endpoint
   end
 
   def load_district
     @district = District.find_by!(name: params[:district_id])
+  end
+
+  def name_prefix
+    "#{@district.name}-"
   end
 end
--- a/app/services/build_heritage.rb
+++ b/app/services/build_heritage.rb
@@ -3,9 +3,9 @@ class BuildHeritage
 
   def initialize(_params, district: nil)
     params = _params.to_h.deep_dup
-    @district = district
     heritage_name = params.delete(:id) || params[:name]
     @heritage = Heritage.find_or_initialize_by(name: heritage_name)
+    @district = district || @heritage.district
     @params = convert_params_for_model params
   end
 
@@ -16,7 +16,16 @@ class BuildHeritage
         service[:port_mappings_attributes] = service.delete(:port_mappings) if service[:port_mappings].present?
 
         unless service[:listeners].nil?
-          listener_map = Hash[Endpoint.where(name: service[:listeners].map { |e| e[:endpoint] }).pluck(:name, :id)]
+          endpoint_names = service[:listeners].map { |e| e[:endpoint] }
+
+          endpoints = if @district.present?
+            Endpoint.where(name: endpoint_names, district: @district).pluck(:name, :id)
+          else
+            Endpoint.where(name: endpoint_names).pluck(:name, :id)
+          end
+
+          listener_map = Hash[endpoints]
+
           service[:listeners_attributes] = service.delete(:listeners).map do |listener|
             {
               endpoint_id: listener_map[listener[:endpoint]],
--- a/app/services/build_heritage.rb
+++ b/app/services/build_heritage.rb
@@ -17,15 +17,9 @@ class BuildHeritage
 
         unless service[:listeners].nil?
           endpoint_names = service[:listeners].map { |e| e[:endpoint] }
-
-          endpoints = if @district.present?
-            Endpoint.where(name: endpoint_names, district: @district).pluck(:name, :id)
-          else
-            Endpoint.where(name: endpoint_names).pluck(:name, :id)
-          end
+          endpoints = Endpoint.where(name: endpoint_names, district: @district).pluck(:name, :id)
 
           listener_map = Hash[endpoints]
-
           service[:listeners_attributes] = service.delete(:listeners).map do |listener|
             {
               endpoint_id: listener_map[listener[:endpoint]],
--- a/spec/requests/create_heritage_spec.rb
+++ b/spec/requests/create_heritage_spec.rb
@@ -3,6 +3,8 @@ require 'rails_helper'
 describe "POST /districts/:district/heritages", type: :request do
   let(:user) { create :user }
   let(:district) { create :district }
+  let(:endpoint) { create :endpoint, district: district }
+
   let(:params) do
     {
       version: version,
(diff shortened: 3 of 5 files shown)


app/controllers/endpoints_controller.rb
```diff
@@ -37,11 +39,26 @@ def permitted_params
     )
   end
 
+  def update_params
```
Yuuki77, line 42: for updating, the passing params are `certificate_id` and `ssl_policy`

https://github.com/degica/barcelona-cli/blob/4509ab33cd2252ff0660a2d5f187767e7d8107b7/cmd/endpoint.go#L159-L163

degicat:
@davidsiaw please review this

app/controllers/endpoints_controller.rb
```diff
@@ -37,11 +39,26 @@ def permitted_params
...
+      :ssl_policy
+    )
+  end
+
   def load_endpoint
-    @endpoint = @district.endpoints.find_by!(name: params[:id])
+    @endpoint = @district.endpoints.find_by(name: params[:id])
+    unless @endpoint.present?
+      @endpoint = @district.endpoints.find_by!(name: name_prefix + params[:id])
+    end
```
davidsiaw, line 53: I think we will need another column for district. This is because its possible that there might exist a district called "food" and a endpoint called "bar". If there was a legacy endpoint called "food-bar", the endpoint names will collide and you will load the wrong endpoint!

To avoid this risk with other users of Barcelona (if any) lets add a column and it will be a lot better.

davidsiaw (changes requested):
@Yuuki77 Lets make a new column for district id on endpoint

app/controllers/endpoints_controller.rb
Yuuki77: You're absolutely right.
I agree with adding the new column.
let me add one

Yuuki77: @davidsiaw  I've been thinking about this issue.
district had already been added to the endpoint

https://github.com/degica/barcelona/blob/2734382417d3670bce5b43676f02c8eca27cb647/db/schema.rb#L53-L63

there is a unique validation for name so at least it should fail to create if there is a existing name which will collide

davidsiaw: The collision can cause other kinds of problems, not just on create, such as an endpoint pointing to the wrong servers. We should eliminate the issue before giving a chance for other issues to crop up.

Yuuki77: If the name is guaranteed to be unique when creating, I don't think it's possible to load the wrong endpoint.

https://github.com/degica/barcelona/blob/07009e6736178a39b36816517af55a766db585c8/app/models/endpoint.rb#L122-L126

When loading, it will look for an endpoint with a matching name in the specified district.
So I dont think it will load wrong one if endpoint is unique.

ok how about using `heritage.id` ?
I think this should decrease the collision with exiting one.

Yuuki77: @davidsiaw 
I added the tests for your concerns https://github.com/degica/barcelona/pull/660/commits/2c5cac8e7328a84e86a485d248cc1d51e6bd9df0

When a collision occurs, the creation of a new one is failing.
To reduce the possibility of a collision with an existing one or one with a district name, we can add an ID to the name, but I think it will be difficult to read the endpoint name when you check it in AWS.
Also, I wonder how much of that collision will happen.
Let me know what you think 🙏

davidsiaw: @Yuuki77 Ah, I don't mean the creation of an endpoint. We talked about that already, I am talking about

```
environments:
  staging:
    name: blue-green
    image_name: quay.io/dsiawdegica/bg
    scheduled_tasks: []
    services:
      - name: web
        service_type: web
        memory: 256
        command: sh /start.sh
        web_container_port: 8080
        listeners:
          - endpoint: staging-blue-green <-------------------------------------
            health_check_path: /

```

Yuuki77: @davidsiaw Thank you for the explanations
now it makes sense to me!
added the filter!
https://github.com/degica/barcelona/pull/660/commits/09477c57ea912d7157dfb6f9af8c4bb713240f23

spec/requests/create_heritage_spec.rb
```diff
@@ -105,5 +110,16 @@
       let(:version) { 2 }
       it_behaves_like "create"
     end
+
+    context "when the endpoint is not belonged to heritage" do
```
davidsiaw, line 114: Is this what you mean?

```suggestion
    context "when the endpoint does not belong to a district " do
```

spec/requests/update_endpoint_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+describe "PATCH /districts/:district/endpoints/:endpoint", type: :request do
```
davidsiaw, line 3: Can we also write a test for the legacy case where the endpoint district is nil?

app/services/build_heritage.rb
```diff
@@ -16,7 +16,16 @@ def convert_params_for_model(new_params)
...
 
         unless service[:listeners].nil?
-          listener_map = Hash[Endpoint.where(name: service[:listeners].map { |e| e[:endpoint] }).pluck(:name, :id)]
+          endpoint_names = service[:listeners].map { |e| e[:endpoint] }
+
+          endpoints = if @district.present?
+            Endpoint.where(name: endpoint_names, district: @district).pluck(:name, :id)
+          else
+            Endpoint.where(name: endpoint_names).pluck(:name, :id)
+          end
```
davidsiaw, line 25: Hmm this would still cause the same issue. Since barcelona is open source if someone that used it made an endpoint called `staging-blue-green` in the past, and now made a new district called `staging-blue` but forgot to make an endpoint called `green` barcelona will choose the old one!

davidsiaw (changes requested):
Needs a few more tests to handle the legacy cases

spec/requests/create_heritage_spec.rb
Yuuki77: ah thanks!

spec/requests/update_endpoint_spec.rb
Yuuki77: @davidsiaw I think `district` cannot be nil when creating an endpoint.
the original code is 
```
  def create
    endpoint = @district.endpoints.create!(permitted_params)
    render json: endpoint
  end
```

app/services/build_heritage.rb
Yuuki77: @davidsiaw 
I think this test shows it works in such a case
https://github.com/degica/barcelona/pull/660/commits/6322a4682ce1a05285048788162111ace32d30c1

app/services/build_heritage.rb
```diff
@@ -16,7 +16,16 @@ def convert_params_for_model(new_params)
         service[:port_mappings_attributes] = service.delete(:port_mappings) if service[:port_mappings].present?
 
         unless service[:listeners].nil?
-          listener_map = Hash[Endpoint.where(name: service[:listeners].map { |e| e[:endpoint] }).pluck(:name, :id)]
+          endpoint_names = service[:listeners].map { |e| e[:endpoint] }
+
+          endpoints = if @district.present?
```
davidsiaw, line 21: @Yuuki77 Hmm. okay, I thought it would be null sometimes when this happens

Yuuki77: @davidsiaw 
in app/controllers/heritages_controller.rb:27
it is passed as nil
```ruby
  def update
    @heritage = BuildHeritage.new(permitted_params).execute
    @heritage.save_and_deploy!(without_before_deploy: false,
                               description: "Update to #{@heritage.image_path}")
    render json: @heritage
  end
```
 but I think since I updated the logic like this in app/services/build_heritage.rb:8
```ruby
    @district = district || @heritage.district
```
district cannot be nil so we can remove this condition.

davidsiaw: How about older data, before your change?

Yuuki77: @davidsiaw 
before it just assigned
`@district = district ` so I thought this can be nil in update.
but `@heritage.district` will not be nil in when heritage already exits

davidsiaw (approved):
LGTM

PR #663, opened 2021-03-23
resource model

davidsiaw:
this is where it starts!


davidsiaw:
Gonna do it another way

PR #664, opened 2021-03-29
Add `get_district` endpoint

Yuuki77:

Currently, it is not possible to retrieve district information from the name of the heritage.
So, add this API to make it happen.

## Note
In this morning's meeting 
I said that the name "heritage" will no longer be unique, but that is not true. it has to be unique
So the district will also be unique.


degicat:
@davidsiaw please review this

Yuuki77:
actually we might not need this endpoint.

PR #667, opened 2021-04-02
Pass district id as parameter for `update` and `trigger`

Yuuki77:

pass district id as parameter for update and trigger endpoints.
in following pr, I will add the logic to create heritage with district name.
By adding this endpoint, we can make the heritage unique for each district in the future.


degicat:
@davidsiaw please review this

app/controllers/heritages_controller.rb
```diff
@@ -28,6 +28,13 @@ def update
     render json: @heritage
   end
 
+  def update_with_district
+    @heritage = BuildHeritage.new(permitted_params, district: @district).execute
+    @heritage.save_and_deploy!(without_before_deploy: false,
+                               description: "Update to #{@heritage.image_path}")
+    render json: @heritage
```
davidsiaw, line 35: can't this just be in `update` but with `district` as an optional?

davidsiaw (changes requested):
Can we use the same update method and just use district if its provided for now?

app/controllers/heritages_controller.rb
Yuuki77: @davidsiaw ok I updated the logic to pass the value in body as optional.
Let me know what you think 🙏

app/controllers/heritages_controller.rb
```diff
@@ -22,7 +23,7 @@ def create
   end
 
   def update
-    @heritage = BuildHeritage.new(permitted_params).execute
+    @heritage = BuildHeritage.new(permitted_params, district: @district).execute
```
davidsiaw, line 26: Thanks this looks like a very clean way to handle it. Its also a very implicit `if` statement, so I think we should have a test to get the heritage for the following cases:

1. endpoint has no district, name passed in
2. endpoint has no district, name and district passed in
3. endpoint has district, name passed in
4. endpoint has district, name and district passed in
5. endpoint has district, with another endpoint in different district with the same name, name passed in
6. endpoint has district, with another endpoint in different district with the same name, name and district passed in

davidsiaw (changes requested):
Looks great! We just need more testing to cover some cases

app/controllers/heritages_controller.rb
Yuuki77: @davidsiaw just to confirm, what do you mean by `endpoint`here?
like `endpoint has no district, name passed in`
this is the case when pass heritage name but there is no district?

davidsiaw: Not pass, but the database has an endpoint but district is set to `nil`

Yuuki77: district has endpionts right?
https://github.com/degica/barcelona/blob/07009e6736178a39b36816517af55a766db585c8/app/models/district.rb#L12-L12

Yuuki77: also district name should be unique
also I could not understand what you mean by 🙏 
```
endpoint has district, with another endpoint in different district with the same name, name passed in
``

Comment (Yuuki77) [reply app/controllers/heritages_controller.rb:26]: @davidsiaw updated the logic
https://github.com/degica/barcelona/pull/667/commits/05f4bae33902f14216b35fc29639dbad47ffb87e
also existing tests cover for when no district name is given

Comment (davidsiaw) [reply app/controllers/heritages_controller.rb:26]: @Yuuki77 I think it is hard to write this test from what I see, because you are testing the API. You should test BuildHeritage instead, and it will be clearer. What we need to test essentially is all the permutations of

1. The various states of the database
2. The inputs passed to update

**Review by davidsiaw — CHANGES_REQUESTED** [review]

File: app/controllers/heritages_controller.rb (hunk 2 above)
Comment (Yuuki77) [reply app/controllers/heritages_controller.rb:26]: @davidsiaw added the test.
I think the other scenario is that the wrong district is passed (which has always been possible).
So I added the checks.
other that existing tests should cover.
The change I made here was to pass `district` in update, which was nil.
 https://github.com/degica/barcelona/pull/667/commits/dd9fef7afc8f901c0d79de894c7b1ec18cbee9a8

File: app/controllers/heritages_controller.rb [hunk 3]
```diff
@@ -151,6 +152,15 @@ def load_district
     @district = District.find_by!(name: params[:district_id])
   end
 
+  def load_district_by_param
+    if params[:district].present?
+      @district = District.find_by!(name: params[:district])
+      if @heritage.district.name != @district.name
+        raise ExceptionHandler::InternalServerError.new("Wrong district was passed")
+      end
+    end
```
davidsiaw, line 161: Nested if and reword the exception

```suggestion
    return if params[:district].empty?
    
    @district = District.find_by!(name: params[:district])
    if @heritage.district.name != @district.name
      raise ExceptionHandler::InternalServerError.new("The heritage #{@heritage.name} does not belong to district #{district.name}")
    end
```

app/services/build_heritage.rb
```diff
@@ -5,6 +5,12 @@ def initialize(_params, district: nil)
     params = _params.to_h.deep_dup
     heritage_name = params.delete(:id) || params[:name]
     @heritage = Heritage.find_or_initialize_by(name: heritage_name)
+
+    # we can remove this check once we find heritage by heritage name and district name
+    if district && @heritage.district && district.id != @heritage.district.id
+      raise ExceptionHandler::InternalServerError.new("Wrong district is provided")
```
davidsiaw, line 11:
```suggestion
      raise ExceptionHandler::InternalServerError.new("#{@heritage.name} does not belong to #{district.name}")
```

spec/requests/update_heritage_spec.rb
```diff
@@ -141,6 +141,86 @@
...
+            command: "rake jobs:work"
+          }
+        ]
+      }
+
+      expect(Heritage.last.services.find_by(name: 'web').cpu).to eq 128
+
+      api_request :patch, "/v1/heritages/nginx", params
+      expect(response.success?).to eq false
+      expect(response.body).to eq  "{\"error\":\"Wrong district was passed\"}"
```
davidsiaw, line 221: use single quotes when you have to use double quotes a lot

```suggestion
      expect(response.body).to eq  '{"error":"Wrong district was passed"}'
```

davidsiaw (changes requested):
Just a few more things

Yuuki77:
Thank you for the reviews 🙇 
I have updated it

davidsiaw (approved):
LGTM

Issue #613, opened 2020-09-03
Do not deploy if the image is unavailable


davidsiaw:
Barcelona should check that a tag exists before attempting to deploy it. For example

```
bcn deploy -e staging --tag nonexistent-tag
```

should return an error

PR #668, opened 2021-04-13
add `retrive-manifest` endpoint

Yuuki77:
part of https://github.com/degica/barcelona/issues/613

Add a ` retrive-manifest` endpoint to Barcerola to resolve this issue.
Initially, I tried to get the manifest using dockercfg, but after checking with quay's CS, he said that was not possible.
So we need a user and password for auth.

We can use this endpoint to check if the docker image is in the registry.

### DIFF
--- a/app/controllers/heritages_controller.rb
+++ b/app/controllers/heritages_controller.rb
@@ -1,6 +1,6 @@
 class HeritagesController < ApplicationController
   before_action :load_district, only:   [:index, :create]
-  before_action :load_heritage, except: [:index, :create, :trigger]
+  before_action :load_heritage, except: [:index, :create, :trigger, :retrieve_manifest]
   skip_before_action :authenticate, only: [:trigger]
 
   def index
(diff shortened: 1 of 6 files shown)


degicat:
@davidsiaw please review this

app/controllers/heritages_controller.rb
```diff
@@ -151,6 +151,20 @@ def load_district
...
+  def retrieve_manifest
+    heritage = Heritage.find_by!(name: params[:heritage_id])
+    user = params[:user]
+    password = params[:password]
+
+    if heritage.image_name.start_with?("quay.io/")
+      response = RetrieveQuayManifest.new().process(user, password, heritage)
+    else
+      raise ExceptionHandler::InternalServerError.new("Unsupported Registry")
+    end
```
davidsiaw, line 163: This is strange because we should be able to login to quay using the docker API. Please see the docker CLI implementation https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/registry/login.go

It is possible for the docker CLI to login to quay without having to use the quay-specific CLI, which I think should be the case with barcelona as well.

davidsiaw (changes requested):
@Yuuki77 Thanks for the PR, I think we can have a closer look at how docker CLI does things to learn how to login to all docker registries universally.

app/controllers/heritages_controller.rb
Yuuki77: @davidsiaw  
Sure, we can use the docker api to `log in`.
but  to use `manifest` endpint `https://docs.docker.com/registry/spec/api/#manifest`
we need bare token  and it requires a `basic auth`
I do not think docker CLI support `manifest` endpoint yet.

```
curl -X GET --user username:password "https://quay.io/v2/auth?service=quay.io&scope=repository:NAMESPACE/REPOSITORY:pull"
```
I think once https://docs.docker.com/engine/reference/commandline/manifest_inspect/ is not experimental,
we can start to look how they are doing it.

davidsiaw (changes requested):
Hmm okay, I think lets put this change aside for now then until we can figure out how to connect to docker registries without having to implement third party apis.

Issue #672, opened 2021-04-20
Delete User record on changing auth-method


essa:
There can be two records for one user when the user change the auth method.

I have used github-auth and logged in with vault-auth. Then there are two records for me.

```
irb(main):008:0> pp User.where(name: 'essa')
D, [2021-04-20T07:11:55.675350 #20] DEBUG -- :   User Load (1.2ms)  SELECT "users".* FROM "users" WHERE "users"."name" = $1  [["name", "essa"]]
[#<User:0x000055f7886d19a8
  id: 2,
  name: "essa",
  token_hash:   "....",
  created_at: Mon, 16 Nov 2015 03:15:55 UTC +00:00,
  updated_at: Tue, 20 Apr 2021 06:19:46 UTC +00:00,
  public_key:   "ecdsa-sha2-nistp521 AAAAE.....",
  roles: ["developer", "admin"],
  auth: "github">,
 #<User:0x000055f7886d1408
  id: 22,
  name: "essa",
  token_hash:   ".....",
  created_at: Thu, 21 Jan 2021 08:24:12 UTC +00:00,
  updated_at: Tue, 20 Apr 2021 06:20:17 UTC +00:00,
  public_key:   "ecdsa-sha2-nistp521 AAAAE......=",
  roles: [],
  auth: "vault">]
```

On auto-refresh, barcelona server tries to update the first record(id=2) to `{ name: 'essa', auth: 'vault' }`.

https://github.com/degica/barcelona/blob/master/app/auth/vault_auth.rb#L14-L24
```ruby
@current_user = User.find_by_token(vault_token)

    if @current_user.nil?
      user = User.find_or_create_by(name: username) # This should find id=22 but finds id=2 instead because name matches
      user.auth = 'vault'
      user.token = vault_token
      user.roles = []
      user.save!
      @current_user = user
    end
```

This produces a Validation error because there is the second record(id=22) which have same name and auth.

So, all the User records with same name should be deleted on changing auth method after validating the token.

PR #673, opened 2021-04-22
Fix to select correct auth user

Yuuki77:
fix https://github.com/degica/barcelona/issues/672

## Context
If a user exists in multiple auth, the wrong user will be selected in the valut auth.

This is  because we do not specify `auth type` in `app/auth/vault_auth.rb`.

If auth is nil, it will be github auth.
so I specify auth when selecting.


degicat:

davidsiaw (approved):
LGTM

Issue #644, opened 2021-01-07
[Proposal] Add secret endpoint


k2nr:
Right now, to create a secret environment variable, we need to manually create SSM parameters on AWS and then specify the SSM parameter path in barcelona.yml

```
environment:
  - name: VARIABLE_NAME
    ssm_path: path/to/ssm
```

I propose a new sub-command `bcn secret` to manage everything:

```
$ bcn secret add my_secret_name <secret value>
```

then in barcelona.yml

```
environment:
  - name: VARIABLE_NAME
    secret_name: my_secret_name
```

when `bcn secret add` is executed, Barcelona adds a new SSM parameter to `/barcelona/<district name>/my_secret_name`

essa:
@k2nr 
Very good idea!

But, is it a subcommand of district command?
if so, command will be something like
```
$ bcn district secret add my_secret_name <secret value> <district name>
```
or 

```
$ bcn secret add my_secret_name <secret value> -d <district name>
```

k2nr:
oh that's a good point. Since the current `ssm_path` prefixes `/barcelona/<district name>`, secret should belong to district.
This way apps in a same district can share secrets, which might be useful.

essa:
It's simple and easy to understand.

- Secrets belongs to a district
- Heritages borrow secrets from its district by declaring in `barcelona.yml`

davidsiaw:
Or we could just have the get/set env modify SSM if the variable is on SSM

PR #675, opened 2021-04-27
Add `get_ssm_parameter` endpoint

Yuuki77:
Part of https://github.com/degica/barcelona/issues/644

Right now, to create a secret environment variable, we need to manually create SSM parameters on AWS and then specify the SSM parameter path in barcelona.yml

we want to automaize this process.
first I added the endpoint to get the `aasm_parameter` to check if the key exists in aws.
after that I will add the endpoint to set the aasm parameter in aws


Gemfile.lock
```diff
@@ -51,17 +51,17 @@ GEM
...
     aws-sdk-autoscaling (1.20.0)
       aws-sdk-core (~> 3, >= 3.48.2)
       aws-sigv4 (~> 1.1)
     aws-sdk-cloudformation (1.18.0)
       aws-sdk-core (~> 3, >= 3.48.2)
       aws-sigv4 (~> 1.1)
-    aws-sdk-core (3.48.3)
-      aws-eventstream (~> 1.0, >= 1.0.2)
-      aws-partitions (~> 1.0)
+    aws-sdk-core (3.114.0)
```
Yuuki77, line 62: I had to update sdk core to use gem 'aws-sdk-ssm'

degicat:

app/controllers/districts_controller.rb
```diff
@@ -50,6 +50,17 @@ def sign_public_key
     render json: json
   end
 
+  def get_ssm_parameter
```
davidsiaw, line 53: Its better to make this a controller `SsmParametersController` and then just use the standard rails actions, then in the route just have 

```
resources :ssm_parameter, only: [:index, :create, :destroy]
```

Have a look at the `HeritageController` on how to do this. This means that your action that you are implementing right now is just `show` and you do not have to explicitly declare it in the routes.rb

app/services/process_ssm.rb
```diff
@@ -0,0 +1,23 @@
+class ProcessSsm
+  def initialize(district, name)
+    @district = district
+    @name = name
+  end
+
+  def get_parameter
```
davidsiaw, line 7: Don't prefix with `get_`, just say `parameter` or `info`

app/services/process_ssm.rb
```diff
@@ -0,0 +1,23 @@
+class ProcessSsm
```
davidsiaw, line 1: Since this class represents an SsmParameter, just call it that.

```suggestion
class SsmParameter
```

config/routes.rb
```diff
@@ -4,6 +4,7 @@
       member do
         post :apply_stack
         post :sign_public_key
+        get "/get_ssm_parameter/:name", to: "districts#get_ssm_parameter"
```
davidsiaw, line 7: see my comment on the controller

davidsiaw (changes requested):
A few things

PR #676, opened 2021-04-28
Endpoint to put ssm parameter

Yuuki77:
~this pr depends on https://github.com/degica/barcelona/pull/675~
Part of #644

now add the endpoint to set ssm parameter to aws
path is `v1/districts/:district_name/ssm_parameter`.

You need to pass 
```ruby
   name: name, # required
   value: value, # required
```


degicat:

app/controllers/create_ssm_parameters_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
```
davidsiaw, line 1: I think you meant to put this file in `spec/controllers` instead?

app/controllers/ssm_parameters_controller.rb
```diff
@@ -0,0 +1,16 @@
+class SsmParametersController < ApplicationController
+  before_action :load_heritage, only: [:create]
```
davidsiaw, line 2: Does this work? There is no `load_heritage` method in this class

spec/services/ssm_parameters_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe SsmParameters do
+  let(:district) { build :district, aws_access_key_id: 'test', aws_secret_access_key: 'test' }
+
+  let(:name) {"PSParameterName"}
+
+  describe "#ssm_path" do
+    it "returns expected ssm_path" do
+      expect(described_class.new(district, name).ssm_path).to eq "/barcelona/#{district.name}/#{name}"
+    end
+  end
```
davidsiaw, line 12: You should use specific strings so its easier to read the test

```suggestion
  let(:district) { build :district, name: 'districtname', aws_access_key_id: 'test', aws_secret_access_key: 'test' }

  describe "#ssm_path" do
    it "returns expected ssm_path" do
      expect(described_class.new(district, 'app/paramname').ssm_path).to eq "/barcelona/districtname/app/paramname"
    end
  end
```

davidsiaw (changes requested):
@Yuuki77 sorry I just got to this 📖 . I have some concerns, if you could address them that would be amazing!

Yuuki77:
@davidsiaw Thank you for the review.
I should have realized before the review.
now it should be ok.

app/services/ssm_parameters.rb
```diff
@@ -0,0 +1,30 @@
...
+  PARAMETER_TYPES = %w[String StringList SecureString].freeze
+
+  def initialize(district, name)
+    @district = district
+    @name = name
+  end
+
+  def put_parameter(value, type)
+    unless PARAMETER_TYPES.include?(type)
+      raise ExceptionHandler::InternalServerError.new("Type #{type} is not in #{PARAMETER_TYPES}")
```
davidsiaw, line 11: Internal Server Errors are errors that are unexpected or temporary and we don't want or don't know what to show the client, and it is not the client's fault. In this case, it is the client's fault for not passing the correct parameter type so this should return something like ParameterError or something to tell the client that what it is passing to the server is not correct and cannot be accepted.

davidsiaw (changes requested):
One more thing, we should return a proper error for a wrong parameter.

There is a further thing that is SSM values that exceed 4096 characters cannot be stored in the normal SSM space, but I think we can make an issue for that.

app/services/ssm_parameters.rb
davidsiaw: Maybe we should just only use SecureString, and remove the parameter, which will clean up all the error handling code.

Yuuki77: @davidsiaw Thank you for the reviews
That makes sense!
I updated the code! 
Can you take a look at when you have time?
https://github.com/degica/barcelona/pull/676/commits/6d7a56b1a809e845bf20ba51c8cc971c5b00117e

Yuuki77:
@davidsiaw also this one 🙏

davidsiaw (approved):
LGTM!

PR #678, opened 2021-05-11
Apply RuboCop

Yuuki77:
In barcelona, rubocop.yml exists, but it is not integrated into ci, so there are places where the rules are not applied.
Therefore, I'd like to run rubocop in Ci.
In order to do this, we must first correct some codes.
this is the first step, I fixed the codes where rubocop can autofix in app folder.
If you think it is better split the code into multiple prs, please let me know.
In next pr, I am going to fix codes in specs.


.rubocop.yml
```diff
@@ -36,10 +34,10 @@ Style/Documentation:
 Style/DoubleNegation:
   Enabled: false
 
-Style/DotPosition:
+Layout/DotPosition:
```
Yuuki77, line 37: go the waring about namespace so fixed those.

.rubocop.yml
```diff
@@ -110,6 +105,52 @@ Style/SingleLineBlockParams:
 Lint/UnderscorePrefixedVariableName:
   Enabled: false
 
+# TODO remove this
```
Yuuki77, line 108: it changes many places so I will remove in later pr

app/middlewares/exception_handler.rb
```diff
@@ -93,8 +93,6 @@ def call(env)
       raise UnprocessableEntity.new(e.message)
     rescue Octokit::ClientError => e
       raise BadRequest.new(e.message)
-    rescue
```
Yuuki77, line 96: this is not used so removed it

degicat:
@Resonious can you help us review this PR, please?

degicat:

app/controllers/heritages_controller.rb
```diff
@@ -84,58 +84,58 @@ def update_service_scale
...
-      ],
-      environment: [
-        :name,
-        :value,
-        :value_from,
-        :ssm_path
-      ]
-    ]).tap do |whitelisted|
+                    :id,
+                    :version,
```
Resonious, line 88: This is insane :laughing:. Should we remove the rule that made this change? Maybe it's no big deal.

app/controllers/review_apps_controller.rb
```diff
@@ -52,11 +52,11 @@ def permitted_params
...
                     :before_deploy,
-                    environment: [
-                      :name,
-                      :value,
-                      :ssm_path,
-                      :value_from
+                    environment: %i[
+                      name
+                      value
+                      ssm_path
```
Resonious, line 58: Seems confusing to mix leading-`:` symbols with `%i[]`-style symbols. I almost thought there we were calling a method called `name`.

Resonious (commented):
Looked through the changes. Mostly straightforward, no problem :+1:. There are a couple of places where the rubocop rules look weird, but it's not a big deal. We can keep them like that.

One question: am I insane or is rubocop not in the `Gemfile`? I think we should put it there, and we should pin the version so that the rules don't suddenly change when we run `bundle install`.

app/controllers/review_apps_controller.rb
Yuuki77: good point!
just disabled it in https://github.com/degica/barcelona/pull/678/commits/b28362c463e81c20aab258e84455ce13b056a236

app/controllers/heritages_controller.rb
Yuuki77: I am not sure how I can disable this 😄 
maybe it is ok but let me know what you think!

Yuuki77:
@Resonious Thank you for the reviews!
as you said, Rubocop only exists in lock file so I added in `Gemfile` as well 🙏

Gemfile
```diff
@@ -12,40 +11,42 @@ gem 'aws-sdk-iam'
...
-  gem 'pry-byebug'
   gem 'debase'
+  gem 'pry-byebug'
+  gem 'pry-rails'
   gem 'ruby-debug-ide'
   gem 'solargraph'
 end
 
 group :development do
+  gem 'rubocop'
```
Resonious, line 38: Personally, I like to set the version for rubocop in the gemfile. Sometimes they update their cops and it breaks randomly.
```suggestion
  gem 'rubocop', '~> 0.86.0'
```

Yuuki77: Thanks! that makes sense!

app/controllers/heritages_controller.rb
davidsiaw: I think you can just move this to a `PERMITTED_PARAMS` constant to avoid the nesting, because some rubocop rules combine to become really funky like this

davidsiaw (changes requested):
Looks mostly good! Just one small thing

app/models/port_mapping.rb
```diff
@@ -50,23 +50,25 @@ def https?
   end
 
   def special_protocol?
-    !%w(tcp udp).include?(protocol)
+    !%w[tcp udp].include?(protocol)
   end
 
   def host_protocol
-    (http? || https?) ? "tcp" : protocol
+    http? || https? ? "tcp" : protocol
```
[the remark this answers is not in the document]
Yuuki77: added one https://github.com/degica/barcelona/pull/678/commits/06b753af11d76961d87bca198980fcff6e0df5a3

Resonious (approved):
LGTM

Yuuki77:
@davidsiaw ping 🙏

davidsiaw (approved):
LGTM

PR #679, opened 2021-05-13
Endpoint remove ssm parameter

Yuuki77:
Part of #644

now add the endpoint to delete ssm parameter from aws
path is `v1/districts/:district_name/ssm_parameter`.

You need to pass 
```ruby
   name: name, # required
```


degicat:

spec/services/ssm_parameters_spec.rb
```diff
@@ -13,9 +13,17 @@
...
       response = described_class.new(district, name).put_parameter(parameter_value)
       expect(response.version).to eq 0
     end
   end
+
+  describe "#delete_parameter" do
+    it "delete ssm parameter" do
+      response = described_class.new(district, name).delete_parameter
+      expect(response.deleted_parameters).to eq []
+      expect(response.invalid_parameters).to eq []
```
davidsiaw, line 26: We should also check here that we are sending out a delete request to AWS. (just mocking AWS will do, we don't have to really send it)

davidsiaw (changes requested):
Looks great! Thank you! Just a small thing

spec/services/ssm_parameters_spec.rb
Yuuki77: @davidsiaw Thank you for the reviews!
updated here https://github.com/degica/barcelona/pull/679/commits/a29189515cec43ee688f371c151b6fb13b1be68c
currently the request is already mocked so it should be ok.

davidsiaw (approved):
LGTM

PR #685, opened 2021-06-10
Fix ssm path for delete endpoint

Yuuki77:
While testing, I noticed that there was a mistake in the delete endpoint.
the name should be ssm_path not `name` itself.
this pr should fix this issue


degicat:

davidsiaw (approved):
LGTM

PR #686, opened 2021-07-09
Validate ssm parameters

Yuuki77:
Currently, when deploying, even if the `ssm_path` does not exist, deploy is performed and cloudformation fails.
adeed the logic to check them before deploying.

### DIFF
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -255,7 +255,7 @@ class Heritage < ActiveRecord::Base
   end
 
   def deploy!(without_before_deploy: false, description: "")
-    validate_ssm_parameters
+    validate_ssm_parameters!
     release = releases.create!(description: description)
     update_services(release, without_before_deploy)
     release
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -255,6 +255,7 @@ class Heritage < ActiveRecord::Base
   end
 
   def deploy!(without_before_deploy: false, description: "")
+    validate_ssm_parameters
     release = releases.create!(description: description)
     update_services(release, without_before_deploy)
     release
@@ -352,5 +353,18 @@ class Heritage < ActiveRecord::Base
 
   def delete_stack
     cf_executor.delete
+  end
+
+  def validate_ssm_parameters
+    ssm_paths = environments.secrets.select{ |key|
+      key.value_from.start_with?("/barcelona/#{district.name}")
+    }.map(&:value_from)
+
+    ssm_parameter = SsmParameters.new(self.district, "")
+    response = ssm_parameter.get_parameters(ssm_paths)
+
+    if response.invalid_parameters.present?
+      raise ExceptionHandler::BadRequest.new("These ssm keys do not exist: #{response.invalid_parameters}")
+    end
   end
 end
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -355,16 +355,16 @@ class Heritage < ActiveRecord::Base
     cf_executor.delete
   end
 
-  def validate_ssm_parameters
+  def validate_ssm_parameters!
     ssm_paths = environments.secrets.select{ |key|
       key.value_from.start_with?("/barcelona/#{district.name}")
     }.map(&:value_from)
 
     ssm_parameter = SsmParameters.new(self.district, "")
-    response = ssm_parameter.get_parameters(ssm_paths)
+    invalid_parameters = ssm_parameter.get_invalid_parameters(ssm_paths)
 
-    if response.invalid_parameters.present?
-      raise ExceptionHandler::BadRequest.new("These ssm keys do not exist: #{response.invalid_parameters}")
+    if invalid_parameters.present?
+      raise ExceptionHandler::BadRequest.new("These ssm keys do not exist: #{invalid_parameters}")
     end
   end
 end
--- a/app/services/ssm_parameters.rb
+++ b/app/services/ssm_parameters.rb
@@ -18,10 +18,11 @@ class SsmParameters
                              })
   end
 
-  def get_parameters(ssm_paths)
-    client.get_parameters({
-                            names: ssm_paths
-                          })
+  def get_invalid_parameters(ssm_paths)
+    response = client.get_parameters({
+                                       names: ssm_paths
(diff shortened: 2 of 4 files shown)


degicat:

app/models/heritage.rb
```diff
@@ -255,6 +255,7 @@ def image_path
   end
 
   def deploy!(without_before_deploy: false, description: "")
+    validate_ssm_parameters
```
davidsiaw, line 258: Add an exclamation mark here since its a method that does not return anything.

```suggestion
    validate_ssm_parameters!
```

app/models/heritage.rb
```diff
@@ -353,4 +354,17 @@ def apply_stack
...
     cf_executor.delete
   end
+
+  def validate_ssm_parameters
+    ssm_paths = environments.secrets.select{ |key|
+      key.value_from.start_with?("/barcelona/#{district.name}")
+    }.map(&:value_from)
+
+    ssm_parameter = SsmParameters.new(self.district, "")
+    response = ssm_parameter.get_parameters(ssm_paths)
```
davidsiaw, line 364: The method `get_parameters` retrieves the parameters for the paths given, so it is a layer of abstraction over the request. This method should not magically know that it made a HTTP Request by assigning the result to a variable named `response`. Instead, `get_parameters` should return the data in a form that is independent of whether or not it is retrieved via HTTP, read from disk, etc...

Please change it so it is possible to mock `get_parameters` with a known return structure or datatype.

app/models/heritage.rb
```diff
@@ -353,4 +354,17 @@ def apply_stack
...
+
+    if response.invalid_parameters.present?
```
davidsiaw, line 366: Owing to the comment above, I think we can just make a `ssm_paths_exist?(ssm_paths)` method on the `SsmParameters` class

app/services/ssm_parameters.rb
```diff
@@ -18,6 +18,12 @@ def delete_parameter
                              })
   end
 
+  def get_parameters(ssm_paths)
+    client.get_parameters({
+                            names: ssm_paths
+                          })
+  end
```
davidsiaw, line 25: This is too raw. Please make this method private and write a method that returns a data type that is independent of the data returned by the AWS SDK.

davidsiaw (changes requested):
Request to clean up a bit of code

Yuuki77:
@davidsiaw Thank you for the reviews!
I updated the logic to just return `invalid params` array
let me know what you think!
https://github.com/degica/barcelona/pull/686/commits/15325358be281ee1ceda5c72b37db7d7940812fa

Yuuki77:
ping @davidsiaw

app/services/ssm_parameters.rb
```diff
@@ -18,6 +18,13 @@ def delete_parameter
                              })
   end
 
+  def get_invalid_parameters(ssm_paths)
+    response = client.get_parameters({
+                                       names: ssm_paths
```
davidsiaw, line 23: @Yuuki77 Sorry it took me so long to re-review this. I found another small issue, if the heritage has no SSM parameters this array will be empty. This causes AWS SDK to throw an error, and results in Barcelona returning a 500 error.

```
F, [2021-07-29T04:08:49.816365 #1] FATAL -- : [7af8f737-d13b-4a03-8044-5bb3fbd38a85] Aws::SSM::Errors::ValidationException (1 validation error detected: Value '[]' at 'names' failed to satisfy constraint: Member must have length greater than or equal to 1.):
```

You can find this error occuring here:  https://ap-northeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-northeast-1#logsV2:log-groups/log-group/Barcelona$252Fdefault$252Fbarcelona/log-events/barcelona$252Fbarcelona-web$252F4fc663b8b6cc45ad9ba42928689294c7

After deploying this branch, I tried to re-deploy but the test server failed to even start deploying due to the error

davidsiaw (changes requested):
Hi @Yuuki77 I found an issue with this code that would prevent Barcelona from deploying, could you please have a look?

app/services/ssm_parameters.rb
Yuuki77: @davidsiaw Thank you for catching.
it looks like it doesn't not throw same error when I call this in test with empty name.
I have updated and please take a look when you have time 🙇 
https://github.com/degica/barcelona/pull/686/commits/922c31ae15b1d9bf7467adfcc467cbb99f851455

app/services/ssm_parameters.rb
```diff
@@ -18,6 +18,15 @@ def delete_parameter
...
   end
 
+  def get_invalid_parameters(ssm_paths)
+    return [] if ssm_paths.empty?
+
+    response = client.get_parameters({
+                                       names: ssm_paths
+                                     })
+    response.invalid_parameters
+  end
```
davidsiaw, line 28: The problem with using outside APIs in such a raw way is that they can throw all sorts of errors you do not care about or throw errors because of parameters that you did not expect they could not handle. I suggest putting a catch-all rescue here and logging the exception.

```suggestion
    response.invalid_parameters
  rescue StandardError => e
    Rails.logger.error("Unexpected error #{e}")
  end
```

davidsiaw (changes requested):
One last thing since any errors that come out from this will prevent deployment and is another thing development will have to query us on.

davidsiaw (changes requested):
One last thing! We may need to make another change to the `bcn` client to pass the exception to the user properly.

app/services/ssm_parameters.rb
```diff
@@ -25,6 +25,8 @@ def get_invalid_parameters(ssm_paths)
                                        names: ssm_paths
                                      })
     response.invalid_parameters
+  rescue StandardError => e
+    Rails.logger.error("Unexpected error #{e}")
```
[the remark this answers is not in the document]
Yuuki77: @davidsiaw 
Sure thing
Updated it here
https://github.com/degica/barcelona/pull/686/commits/c0e4eab9501df14c232a3924d6a5c4b039618b7c

davidsiaw (approved):
Thanks! LGTM

PR #690, opened 2021-08-04
Validate ssm parameters

Yuuki77:
Original pr was reverted https://github.com/degica/barcelona/pull/689

Currently, when deploying, even if the ssm_path does not exist, deploy is performed and cloudformation fails.
added the logic to check them before deploying.

 we seem to have an issue with too many parameters to list that we did not anticipate.
so in this pr, I added the logic to handle in that case

### DIFF
--- a/app/models/heritage.rb
+++ b/app/models/heritage.rb
@@ -255,6 +255,8 @@ class Heritage < ActiveRecord::Base
   end
 
   def deploy!(without_before_deploy: false, description: "")
+    validate_ssm_parameters!
+
     release = releases.create!(description: description)
     update_services(release, without_before_deploy)
     release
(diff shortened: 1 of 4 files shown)


app/models/heritage.rb
```diff
@@ -353,4 +353,19 @@ def apply_stack
   def delete_stack
     cf_executor.delete
   end
+
+  def validate_ssm_parameters!
+    ssm_path_array = environments.secrets.select{ |key|
+      key.value_from.start_with?("/barcelona/#{district.name}")
+    }.map(&:value_from).each_slice(10).to_a
```
Yuuki77, line 360: Divide the array into arrays of length 10.

degicat:

davidsiaw (approved):
LGTM

PR #693, opened 2021-09-01
Revert #660 Fix endpoint is global

Yuuki77:
Revert #660 
When I talked to David, we decided to revert this change

since it will add district name and it causes the barcelona.yml files to break


degicat:

davidsiaw (approved):
LGTM

PR #695, opened 2021-09-27
Add error message and update instance class to a usable one

davidsiaw:
This PR has a few changes:

1. It updates the default allocated instance type to a new one for new versions of Postgres
2. Use the `<<~` syntax to avoid having to indent inline code weirdly.
3. Return the stack status as well to ease diagnosis


degicat:
@FTLam11 can you help us review this PR, please?

FTLam11 (approved):
L:zany_face:TM, thanks for this PR! Squiggly heredoc rocks 🪨

PR #702, opened 2021-10-22
Add workflows for dockerfiles

davidsiaw:
Previously we relied on automated builds on Quay to do this work. Now that we are moving to ECR, we will need to do this ourselves. This PR introduces build instructions for our dockerfiles.

We never wrote tests for these so I will add them in a follow up PR


rramsden (approved):
@davidsiaw Looks good to me. Would be good to reduce the boilerplate with a custom action in the future but not urgent.

PR #703, opened 2021-10-28
Check if the image exists in ecr

Yuuki77:
now added the feature to check if the image exists in ecr

### DIFF
--- a/app/services/ecr_service.rb
+++ b/app/services/ecr_service.rb
@@ -40,6 +40,6 @@ class EcrService
   end
 
   def public_ecr?
-    @heritage.image_name.match(/^(public.ecr.aws)/)
+    @heritage.image_name.match(/^public\.ecr\.aws/)
   end
 end
--- a/app/services/ecr_service.rb
+++ b/app/services/ecr_service.rb
@@ -25,23 +25,7 @@ class EcrService
   private
 
   def ecr
-    @ecr ||= if public_ecr?
-               ecr_public
-             else
-               ecr_private
-             end
-  end
-
-  def ecr_public
-    @heritage.district.aws.public_ecr
-  end
-
-  def ecr_private
-    @heritage.district.aws.ecr
-  end
-
-  def public_ecr?
-    @heritage.image_name.match(/^public\.ecr\.aws/)
+    @ecr ||= @heritage.district.aws.ecr(@heritage.image_name)
   end
 
   # The string after the last / will be matched.
--- a/app/services/ecr_service.rb
+++ b/app/services/ecr_service.rb
@@ -1,6 +1,10 @@
 class EcrService
   def validate_image!(heritage)
     @heritage = heritage
+
+    # The string after the last / will be matched.
+    # For example, when the Image name is public.ecr.aws/degica/barcelona,
+    # it will return barcelona.
     repository_name = @heritage.image_name[%r{/([^/]*?)$}, 1]
 
     begin
--- a/app/services/ecr_service.rb
+++ b/app/services/ecr_service.rb
@@ -4,22 +4,16 @@ class EcrService
   end
 
   def validate_image!
-    begin
-      response = ecr.describe_images({
-                                       image_ids: [
-                                         {
-                                           image_tag: @heritage.tag
-                                         }
-                                       ],
-                                       repository_name: repository_name
-                                     })
-    rescue Aws::ECR::Errors::RepositoryNotFoundException,
-           Aws::ECRPublic::Errors::RepositoryNotFoundException
-      raise ExceptionHandler::BadRequest.new("Image not found in ECR: #{@heritage.image_path}")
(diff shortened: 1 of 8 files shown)


degicat:

app/services/ecr_service.rb
```diff
@@ -0,0 +1,45 @@
...
+  def ecr_public
+    @heritage.district.aws.public_ecr
+  end
+
+  def ecr_private
+    @heritage.district.aws.ecr
+  end
+
+  def public_ecr?
+    @heritage.image_name.match(/^(public.ecr.aws)/)
```
davidsiaw, line 43: `.` matches every character in a regex, so in your regex its possible to match a repository called `quay.io/degica/publicyecrxaws`

```suggestion
    @heritage.image_name.match(/^public\.ecr\.aws/)
```

app/services/ecr_service.rb
```diff
@@ -0,0 +1,45 @@
+class EcrService
+  def validate_image!(heritage)
+    @heritage = heritage
+    repository_name = @heritage.image_name[%r{/([^/]*?)$}, 1]
```
davidsiaw, line 4: what does this do?

davidsiaw (changes requested):
A few things. Will test it in a bit

app/services/ecr_service.rb
Yuuki77: Is that true?
Even with the current regex, I don't think `quay.io/degica/publicyecrxaws` matches.
https://regex101.com/r/cyiNMR/1
but yeah It is better to escape `.` so I will take your suggestion.

app/services/ecr_service.rb
Yuuki77: added the comments
let me know if this is still unclear.

Yuuki77:
Thank you for the reviews!
I have updated the pr!

app/services/ecr_service.rb
```diff
@@ -0,0 +1,49 @@
+class EcrService
+  def validate_image!(heritage)
+    @heritage = heritage
```
davidsiaw, line 3: Please initialize the instance variable in a constructor

```suggestion
  def initialize(heritage)
    @heritage = heritage
  end

  def validate_image!(heritage)
```

app/services/ecr_service.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    # The string after the last / will be matched.
+    # For example, when the Image name is public.ecr.aws/degica/barcelona,
+    # it will return barcelona.
+    repository_name = @heritage.image_name[%r{/([^/]*?)$}, 1]
```
davidsiaw, line 8: Thanks for the explanation. This comment is a test case so you can actually make this a method and test it.

```
  def repository_name
    @heritage.image_name[%r{/([^/]*?)$}, 1]
```

```
  ecrservice.new(Heritage.new(image_name:'public.ecr.aws/degica/barcelona'))
  expect(ecrservice.repository_name).to eq 'barcelona'
```

Also `*` is a zero-or-more matcher so there is no need to do a non-greedy `*?`. Just go `+` will be fine. `%r{/([^/]+)$}`

davidsiaw (changes requested):
Please make the comment an executed test case

app/services/ecr_service.rb
Yuuki77: thanks!
updated it https://github.com/degica/barcelona/pull/703/commits/3cd0291f8e4148297ed33033d1835bf438bea9ff

app/services/ecr_service.rb
```diff
@@ -0,0 +1,53 @@
...
+             end
+  end
+
+  def ecr_public
+    @heritage.district.aws.public_ecr
+  end
+
+  def ecr_private
+    @heritage.district.aws.ecr
+  end
```
davidsiaw, line 41: I htink you can move this into the district class, instead of using the `aws` property directly inside

```suggestion
  @ecr ||= heritage.district.ecr
```

davidsiaw (changes requested):
Another small thing, to clean it up a bit

app/services/ecr_service.rb
Yuuki77: updated it https://github.com/degica/barcelona/pull/703/commits/67214810305cc5efe68a46d39e4de45ef60098e5

davidsiaw:
I gave this a test on our test barcelona

スクリーンショット 2021-11-18 19 11 12

I checked the error and it turns out there are some permissions errors. Do you think:

1. Can we forward these errors back to the user so they don't have to check the barcelona logs just to find that their barcelona IAM has no access to the repo?
2. This is a public repo so it should just be able to check for images normally.

スクリーンショット 2021-11-18 19 12 53

davidsiaw (changes requested):
See comment

davidsiaw:
closing. maintainer has left the company.

PR #713, opened 2022-02-25
Relax stack status check frequency

k2nr:
When there are many apps (normal apps and reviewapps), the DescribeStacks API call gets rate-limited. To mitigate the issue, we can loosen the service status check frequency from 5 seconds to 60 seconds, and heritage stack status check from 5 seconds to 30 seconds. This way deployment time-to-complete will get longer up to 30 seconds, and the deployment complete notification can be delayed up to 55 seconds. I think it's acceptable and much better than deployments being rejected ☺️


degicat:

davidsiaw (approved):
That is crazy, I did not realize this was causing the rate limit issue. Thanks. LGTM

PR #723, opened 2022-05-20
Better branch name error

davidsiaw:
Right now if you deploy a reviewapp with the branch name containing big letters it will fail. This PR will make the message better so the user can fix the problem without having to inspect logs and trace the error back to the source where the verification occurs.


degicat:
@pmq20 can you help us review this PR, please?

Metallion:
Thank you! ^_^

pmq20 (approved):
Thanks! LGTM

PR #726, opened 2022-05-27
Yet another attempt to reduce describe stacks

k2nr:
- `MonitorDeploymentJob` runs 1 minute later. checking status right after deployment trigger doesn't make sense since deploying a service typically takes at least 1 minute
- `sleep 3`  after triggering service deployment
  - `service.deploy` calls `DescribeStacks`


degicat:

davidsiaw (approved):
Tested it and it works. LGTM

PR #606, opened 2020-08-05
Expose status of deployment to the api

davidsiaw:
# Context

Before it was not possible to use the API to find out if a deployment was done yet. This PR fixes this. It adds a `deployed` key to the heritage output and each of its services.

# How to test

1. Deploy a heritage (lets call it `hello`)
2. `bcn api get /heritages/hello`
3. Look for a `deployed` key in the output JSON.

スクリーンショット 2020-08-05 17 52 31

4. Wait for the deployment to finish
5. Run the command again and expect true to be returned this time.

PR #730, opened 2022-06-10
Revert "Merge pull request #606 from degica/expose-deployed"

k2nr:
This reverts commit da1b113cc9458bcb35f0e99f6d2b5c2d66b76073, reversing
changes made to 38e7bfc7e2812563c787a531acfbe33140b4ca85.

these `deployed` parameter calls CloudFormation's `DescribeStacks` API, 2 times per service so if a heritage has 6 services the API is called 12 times which exceeds CF's rate limit.
Currently the parameter is not actively used so let's revert it. if in the future we need to check deployment status, I think we should add a dedicated API for that.


degicat:

davidsiaw (approved):
LGTM

PR #733, opened 2022-07-07
Update service count in place

davidsiaw:
Previously we used the heritage deploy service to set the task count desired. The issue with this was that the heritage deployer would always create a new task definition, and then deploy that. This meant that ECS would create a new set of tasks with the new count before removing the old set of tasks. This resulted in if you wanted to scale a service from 10 to 20 tasks, you would have a transient situation of 30 tasks online.

This was suboptimal, and this PR fixes that. It calls ECS directly to modify the task count, and updates the db so non-changes do not get propagated.

## How to test

You can use a local ver to test it on any account you have
1. `bcn login https://mytestbarcelona`
2. `bcn district create --region ap-northeast-1 test`
3. `bcn endpoint create -d test --public hello`
4. `cd barcelona-hello` <- this is just the barcelona hello repo
5. `bcn create -e hello -d test`
6. `bcn api post /districts/test/heritages/hello/services/web/count '{"desired_container_count": 2}'`

note that no new task definition is created but the container count is immediately updated.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -123,7 +123,7 @@ class Service < ActiveRecord::Base
   private
 
   def ecs
-    ecs = district.aws.ecs
+    @ecs ||= district.aws.ecs
   end
 
   def validate_listener_count
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -96,7 +96,16 @@ class Service < ActiveRecord::Base
     })
   end
 
+  def service_exists
+    !arn.nil?
+  end
+
+  class ServiceNotFoundException < StandardError
+  end
+
   def logical_name
+    raise ServiceNotFoundException, arn_prefix unless service_exists
+
     arn.split('/').last
   end
 
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -85,7 +85,46 @@ class Service < ActiveRecord::Base
     backend.deployment_finished?(deployment_id)
   end
 
+  def save_and_update_container_count!(desired_container_count)
+    self.desired_container_count = desired_container_count
+    self.save!
+
+    ecs.update_service({
+      desired_count: desired_container_count,
+      cluster: district.name,
+      service: logical_name
+    })
+  end
+
+  def logical_name
+    arn.split('/').last
+  end
+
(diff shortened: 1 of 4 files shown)


degicat:

degicat:
@kenta-s can you help us review this PR, please?

app/models/service.rb
```diff
@@ -85,8 +85,47 @@ def deployment_finished?(deployment_id)
...
+      service: logical_name
+    })
+  end
+
+  def logical_name
+    arn.split('/').last
+  end
+
+  def arn
+    service_arns.find { |x| x.start_with? arn_prefix }
```
kenta-s, line 104: Question: do `service_arns` always have at least one arn starting with the prefix? Otherwise `logical_name` can raise `NoMethodError`

kenta-s (commented):
Just one question from me

app/models/service.rb
```diff
@@ -85,8 +85,47 @@ def deployment_finished?(deployment_id)
...
+  def save_and_update_container_count!(desired_container_count)
+    self.desired_container_count = desired_container_count
+    self.save!
+
+    ecs.update_service({
+      desired_count: desired_container_count,
+      cluster: district.name,
+      service: logical_name
+    })
+  end
```
essa, line 97: This will fail if `desired_container_count` is greater than the max of the service.
I think it's better to check the result and raise an exception on failure.

It's ok if you can detect it by the output of `bcn` command.

essa (approved):
LGTM 👍 besides my comments.

We can test it for a while and refine later.

davidsiaw:
@essa @kenta-s I added some error handling. Please have another look. Thanks.

app/models/service.rb
davidsiaw: スクリーンショット 2022-07-08 10 58 53
 
I attempted to find the max count but I could not. I am not sure it exists

essa: Ah, I confused it with ASG, ECS service doesn't have max.

kenta-s (approved):
Thanks! LGTM 👍

PR #738, opened 2022-08-10
Specify a better DeploymentConfiguration on deployment

essa:
This PR specifies `MaximumPercent` as 150 instead of 200, the default value on deployment.

When you set `desired_count` 100 and default, ECS tries to invoke 100 tasks at once, which causes rate limit error.

Specifying  `MaximumPercent` as 150 makes ECS to invoke 50 tasks and wait for them to be deployed before invoking others.

### DIFF
--- a/app/models/backend/ecs/v2/service_stack.rb
+++ b/app/models/backend/ecs/v2/service_stack.rb
@@ -1,11 +1,13 @@
 module Backend::Ecs::V2
   class ServiceStack < CloudFormation::Stack
     class Builder < CloudFormation::Builder
+      DEPLOYMENT_CONFIGURATION = { "MaximumPercent"=>150, "MinimumHealthyPercent"=>100 }
       def build_resources
         add_resource("AWS::ECS::Service", "ECSService") do |j|
           j.Cluster district.name
           j.TaskDefinition options[:task_definition]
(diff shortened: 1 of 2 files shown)


degicat:

app/models/backend/ecs/v2/service_stack.rb
```diff
@@ -1,11 +1,13 @@
 module Backend::Ecs::V2
   class ServiceStack < CloudFormation::Stack
     class Builder < CloudFormation::Builder
+      DEPLOYMENT_CONFIGURATION = { "MaximumPercent"=>150, "MinimumHealthyPercent"=>100 }
```
davidsiaw, line 4: Do you think it will be better to set this to a configurable value, so people can decide if they want lower rolling rate? Perhaps add a variable to the model, but default to 200?

davidsiaw (commented):
@essa Since a change like this changes every single project it depends on, maybe we should default to the old behavior and make it possible for people to set the number as they go as they need. People who currently use barcelona might be surprised by such a wide-reaching change

davidsiaw:
We had a closer look and decided that a 150% change would cause a slowdown only in deployment and **shouldn't** create issues. We will make this a more opt in change later down the road.

PR #740, opened 2022-08-22
Fix internal server error

davidsiaw:
There was an edge case where if your token expires barcelona can throw a 403 error during the initialization of the Vault client. This results in bcn returning an Internal server error.

This PR fixes that, and returns the appropriate error with a more legible message instead.

It is a big pain to set up but you can
1. use a/the test vault server and barcelona server
2. make a vault token and successfully log in
3. delete it on vault
4. try log in again and see that you receive 403 instead of internal server error


degicat:
@Yuuki77 can you help us review this PR, please?

Yuuki77 (approved):
LGTM!

PR #743, opened 2022-10-07
Remove NAT ami image listing

essa:
We are using Nat Gateway and don't use NAT instances now, so this part is not needed anymore.


degicat:
@ochko can you help us review this PR, please?

ochko (approved):
Thank you for the cleanup.
LGTM!

PR #757, opened 2023-02-15
Fix Docker build issues

matafc:
The first time I tried to bring up Barcelona on my laptop, local builds and docker builds did not complete. I decided to focus on the fixing the docker builds. Here are the solutions.
- Updated the Gemfile.lock with more recent version. This should also resolve some dependabot issues like https://github.com/degica/barcelona/pull/756
- Updated buildpacks at https://github.com/degica/dockerfiles/pull/15, including adding 2.7.7 ruby. This was required due to some issues with building some gems on the old 2.6 image(at least locally). Using a more recent ruby base to build from resolved them, I think part of the issue was out of date OS dependencies not working when I updated the gems.
- Upgraded image ruby to 2.7.7
- Use platform flag when buliding docker images, and when running app with `kaiser`. Without the correct arch set, on M1 you will get errors like: 

```
Could not find racc-1.6.2 in any of the sources
Run `bundle install` to install missing gems.
```
```
docker: Error response from daemon: pull access denied for kaiser, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```
- Fix Makefile and docker-compose `/bin/sh: UID: readonly variable` was a problem with me useing zsh, and also added platform env for the docker build/run
- Updated Github Actions workflows to remove [deprecation warnings](https://github.com/degica/barcelona/actions/runs/4142080001)


degikko:
@Agroang can you help us review this PR, please?

Agroang (commented):
Sorry @matafc , I am not really sure how to test this. Should I just try to run barcelona within this branch?
Any instructions would be welcomed as I am also not very familiar with this repo 🙇

matafc:
No problem, I'm figuring that out as well.

davidsiaw:
(of course you also want to check that you are on this branch)

matafc:
I just rebased to include the latest updated master. This is ready to test locally now. @davidsiaw should we test this in staging somehow before merging?

davidsiaw:
@matafc  If you start the thing up in Kaiser, and make a district on the sandbox and deploy the hello world successfully then we should be golden.

Agroang (approved):
I definitely lack the knowledge to say that this PR LGTM but at least I can say that the recommended test by David worked as expected (running `docker build -t barcelona .` within the branch). 
I won't merge as I believe you guys should have more knowledge when it comes to the right timing.

Issue #762, opened 2023-03-14
Failure of draining instances on scale down


essa:
When you scale container instances more than 100, there are two cases of failing to set the instance status as DRAINING. This causes instance shutdown which is ECS tasks are running.

- ThrottlingException
- More than 100 container instances

### Background

Barcelona set lifecycle hook, [drain_instance.py](https://github.com/degica/barcelona/blob/7dbeb44dafdfde7381984619501eda5217891218/drain_instance.py) on termination of instances. It is responsible for let ASG wait for the termination of ECS tasks.

1. It is invoked with EC2 instance id
2. Get the ECS container instance ARN from the EC2 instance id
3. Check any ECS tasks are running on the instance
4. If there are no task, complete the lifecycle
5. If there are tasks running, sleep some time and post the trigger to redo the process

### ThrottlingException

This hook is called once per instance. So, if there are many instances, this hook is called many times and call AWS api many times.

I'm not sure about the threshold, but I have confirmed following log.

```
An error occurred (ThrottlingException) when calling the DescribeContainerInstances operation (reached max retries: 4): Rate exceeded: ClientError
Traceback (most recent call last):
  File "/var/task/index.py", line 31, in lambda_handler
    ciId, status = ciFor(ec2Id)
  File "/var/task/index.py", line 13, in ciFor
    descResp = ecs.describe_container_instances(cluster=clusterName, containerInstances=listResp['containerInstanceArns'])
  File "/var/runtime/botocore/client.py", line 253, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/var/runtime/botocore/client.py", line 557, in _make_api_call
    raise error_class(parsed_response, operation_name)
ClientError: An error occurred (ThrottlingException) when calling the DescribeContainerInstances operation (reached max retries: 4): Rate exceeded
```

This happens on various kind of APIs

### Lack of pagination

The ListContainerInstances API has a pagination. If there are more results than `max_result` which is default to 100, you need to call it again with the token given for continuetion.

Current implementation doesn't have this feature

### What happens after failing Draining

ASG waits for the call of [update_container_instances_state](https://github.com/degica/barcelona/blob/7dbeb44dafdfde7381984619501eda5217891218/drain_instance.py#L36) for one hour, then it will force to continue the terminating process.

From ECS point of view, this means that a ACTIVE instance terminates and the ECS tasks on it will be killed suddenly.

PR #763, opened 2023-03-14
Update drain_instance.py for more than 100 instances and ThrottlingExceptions

essa:
Fixes https://github.com/degica/barcelona/issues/762

This PR also add a random wait for reducing the chance of ThrottlingException.

I have tested it on staging district by scaling up to 150 and down to 10 sometimes. All the container instances are draind.

### DIFF
--- a/drain_instance.py
+++ b/drain_instance.py
@@ -2,6 +2,7 @@ import boto3
 import json
 import os
 import time
+import random
 
 session = boto3.session.Session()
 ecs = session.client(service_name='ecs')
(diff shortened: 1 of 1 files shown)


degikko:
@Maduranga can you help us review this PR, please?

drain_instance.py
```diff
@@ -28,16 +29,24 @@ def lambda_handler(event, context):
...
+
+    try:
+        ciId, status = ciFor(ec2Id)
+        if ciId == None:
+            return
+        if status != 'DRAINING':
+            ecs.update_container_instances_state(cluster=clusterName,containerInstances=[ciId],status='DRAINING')
+
+        tasks = ecs.list_tasks(cluster=clusterName, containerInstance=ciId)['taskArns']
+        if len(tasks) > 0:
```
Maduranga, line 44: @essa 
Not sure if I misunderstood the algorithm.
Here, we are checking if the container instance (EC2 instance) still have any tasks running.
If `len(tasks)>0`, there are still tasks running, so we wait for 5 seconds.
After the 5 seconds, shouldn't we again do `tasks = ecs.list_tasks(...)` and check `len(tasks)>0`?
Shouldn't we have a `while len(tasks) > 0 ...`?

essa: > After the 5 seconds, shouldn't we again do tasks = ecs.list_tasks(...) and check len(tasks)>0?

Yes, and we are doing exactly what you are saying from the beginning of Barcelona. But we do it by triggering a call for this Lambda Function.

```python
 session.client('sns').publish(TopicArn=topicArn, Message=json.dumps(msg), Subject='Invoking lambda again')
```

This for reducing the execution time of Lambda. Stopping tasks will take five to ten minutes sometimes.

Maduranga: Understood.

Maduranga (approved):
LGTM.

PR #767, opened 2023-03-28
Update to gp3 block store

davidsiaw:
Amazon's GP3 block stores are more customizable and use less money. We can update this for our own purposes.


davidsiaw:
superseded by https://github.com/degica/barcelona/pull/789

PR #771, opened 2023-04-19
Adding alb trace id to the nginx log

Maduranga:
This PR add the AWS ALB `trace-id` to the Nginx log.
This helps us in tracking and correlating ALB logs with Nginx logs.


essa:
LGTM 👍

PR #774, opened 2023-05-09
Count STOPPING state ECS tasks

### LINKED ISSUE #772: Lifecycle hook doesn't wait for termination of tasks
Lifecycle hook waits for termination of all the tasks [here](https://github.com/degica/barcelona/blob/master/drain_instance.py#L51).

But it is counting only RUNNING tasks.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs/client/list_tasks.html
> The default status filter is RUNNING

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-lifecycle.html 
ECS Tasks becomes Stopping state before its termination.

So the lifecycle hook allows instance termination even if there are Stopping tasks which are processing requests.

essa:
Resolves #772 

And fixed error handling.

### DIFF
--- a/drain_instance.py
+++ b/drain_instance.py
@@ -3,7 +3,9 @@ import json
 import os
 import time
 import random
+
 from botocore.config import Config
+from botocore.exceptions import ClientError
 
 session = boto3.session.Session()
 config = Config(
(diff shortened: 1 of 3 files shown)


degikko:
@Maduranga can you help us review this PR, please?

Maduranga (approved):
LGTM.
Thank you.

PR #781, opened 2023-05-29
rails7 and ruby3

matafc:
Details

- Upgrade to rails 7
- Upgrade to ruby 3.2.2
- Remove debase and ruby-debug-ide due to https://github.com/ruby-debug/debase/issues/76, fixing that can be in scope of another PR if anyone wants it
- Some syntax and config changes due to updated dependencies
- hosts.config needs to be defined now
- updated docker-compose mounts to make iteration easier without have to rebuild the images

This was tested locally with the [e2e](https://github.com/degica/barcelona-e2e) test.


degikko:

config/environments/production.rb
```diff
@@ -83,4 +83,7 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.hosts << "barcelona.degica.com"
+  
 end
```
essa, line 89: I think you have to exclude a path for ALB health check.
Similar issue is [here](https://github.com/degica/hcs/pull/501).

essa (changes requested):
Almost perfect besides one issue about ALB health check.
See my comment for detail

config/environments/production.rb
matafc: Thank you for that! I was worried I was missing something else. Updated to include the fix.

PR #782, opened 2023-06-01
Revert "rails7 and ruby3"

matafc:
Reverts degica/barcelona#781


app/models/oneoff.rb
```diff
@@ -84,8 +84,8 @@ def stopped?
     %w[STOPPED MISSING].include?(status)
   end
 
-  def run!(**args)
-    run(**args)
+  def run!(*args)
+    run(*args)
```
davidsiaw, line 88: I can understand the change from `(**options)` to `(options)` due to the change in 3.2 on keyword args and hash, but these mean two different things. I would be surprised if it actually works.

davidsiaw (approved):
Looks good, except one thing I am suspicious of.

app/models/oneoff.rb
matafc: I was not sure, but it made the unit tests pass. I still am not sure of all the features barcelona has that could break, so pushing this out was too optimistic for now.

PR #789, opened 2023-07-05
Instance type M6 and Volume gp3, AMI update to Amazon Linux 2 2.0.20230628.

matafc:
## Changes
- Updates the volume type for the bastion and container instances to gp3, from gp2.
- Defines m6 type instances as EBS optimized, per https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html . This will allow us to update the instances to m6i 
- Sets the MaxBatchSize for AutoscalingRollingUpdate to 20. This should reduce `resource contention` or `API Rate Limit exceeded` type errors when applying Cloudformation changes.
- Also including the latest AMI update so our instances will remain compliant. Details at the section below

This has been tested on a local instance of Barcelona with the sre-sandbox AWS account.

### How to roll out

We can update the instance type and volume type at the same time. Both changes require a `bcn district apply `, and Cloudformation change set execution. These steps will allow you to only perform it once.

- Merge and confirm that this PR's changes has been deployed to Barcelona.
- `bcn api patch /districts/:district_name '{ "cluster_instance_type": "m6i.large"}'` # replace `:district_name`
- `bcn district apply :district_name`
- Review and execute the Cloudformation changeset that was produced.

## AMI Update

This PR will also update the AMIs of the container instances as described in this [guide](https://www.notion.so/How-to-Update-AMI-for-komoju-district-6980c4e127774d9791d44d9ad51b0321):

[Release notes](https://docs.aws.amazon.com/AL2/latest/relnotes/relnotes-20230629.html)

[Amazon ECS-optimized AMIs - Amazon Elastic Container Service](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html)

ECS AMI SSM path: https://ap-northeast-1.console.aws.amazon.com/systems-manager/parameters/aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id/description?region=ap-northeast-1
https://ap-northeast-1.console.aws.amazon.com/systems-manager/parameters/aws/service/ecs/optimized-ami/amazon-linux-2/recommended/description?region=ap-northeast-1

And it also updates the AMI for the bastion image.

Bastion AMI SSM path: https://ap-northeast-1.console.aws.amazon.com/systems-manager/parameters/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2/description?region=ap-northeast-1

PR #791, opened 2023-07-24
Fix regex for ebs optimized setting

matafc:
Fix for https://github.com/degica/barcelona/pull/789

https://github.com/degica/barcelona/pull/789 has been deployed, but not applied to any districts yet. We will deploy this fix and then proceed with updating the districts.

It seems that all newer instance types support this, we just want to make sure this flag is enabled for m6 types in the Cloudformation when updating.
This should match any m6 instance type defined at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-optimized.html#ebs-optimization-support

### DIFF
--- a/lib/barcelona/network/autoscaling_builder.rb
+++ b/lib/barcelona/network/autoscaling_builder.rb
@@ -26,7 +26,7 @@ module Barcelona
 
       def ebs_optimized_by_default?
         # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
-        !!(instance_type =~ /\A(a1|c4|c5.?|d2|f1|g3.?|h1|i3|m4|m5.?|m6()?|p2|p3(dn)?|r4|r5.?|t3|u-.*|x1.?|z1d)\..*\z/)
+        !!(instance_type =~ /\A(a1|c4|c5.?|d2|f1|g3.?|h1|i3|m4|m5.?|m6.+||p2|p3(dn)?|r4|r5.?|t3|u-.*|x1.?|z1d)\..*\z/)
       end
 
       def build_resources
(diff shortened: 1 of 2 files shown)


degikko:

lib/barcelona/network/autoscaling_builder.rb
```diff
@@ -26,7 +26,7 @@ class AutoscalingBuilder < CloudFormation::Builder
 
       def ebs_optimized_by_default?
         # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
-        !!(instance_type =~ /\A(a1|c4|c5.?|d2|f1|g3.?|h1|i3|m4|m5.?|m6()?|p2|p3(dn)?|r4|r5.?|t3|u-.*|x1.?|z1d)\..*\z/)
+        !!(instance_type =~ /\A(a1|c4|c5.?|d2|f1|g3.?|h1|i3|m4|m5.?|m6.+||p2|p3(dn)?|r4|r5.?|t3|u-.*|x1.?|z1d)\..*\z/)
```
davidsiaw, line 29: might be an excess `|` character here

davidsiaw (changes requested):
@matafc can we have a simple test that exercises this change? 

```regex
allow(it).to receive(:instance_type) { 'm6i.large' }
expect(it).to be_ebs_optimized_by_default
```

lib/barcelona/network/autoscaling_builder.rb
matafc: Ugh, thanks again.

matafc: Added some tests for this now.

davidsiaw (approved):
LGTM

PR #794, opened 2023-08-18
Use the SDK V3 Style

davidsiaw:
The new node version uses SDK V3. This comes with a new import method.

### DIFF
--- a/barcelona.yml
+++ b/barcelona.yml
@@ -26,10 +26,7 @@ environments:
     name: barcelona
     image_name: public.ecr.aws/degica/barcelona
     before_deploy: rake db:migrate
-    scheduled_tasks:
-      # 10AM JST every week day
-      - schedule: cron(0 1 ? * MON-FRI *)
-        command: bin/chaos
+    scheduled_tasks: []
     services:
       - name: web
         service_type: web
@@ -45,3 +42,4 @@ environments:
         command: rake jobs:work
         cpu: 128
         memory: 256
+ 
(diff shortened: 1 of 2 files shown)


schedule_handler.js
```diff
@@ -1,11 +1,12 @@
-var AWS = require('aws-sdk');
+import { ECSClient, RunTaskCommand } from "@aws-sdk/client-ecs";
+
 var util = require('util');
 
 exports.handler = function(input, context) {
```
Resonious, line 5:
```suggestion
exports.handler = async function(input, context) {
```

Since there's an `await` down below, we'll need this to be `async`.

davidsiaw: Nice catch. Thanks

degikko:

barcelona.yml
```diff
@@ -26,10 +26,7 @@ environments:
     name: barcelona
     image_name: public.ecr.aws/degica/barcelona
     before_deploy: rake db:migrate
-    scheduled_tasks:
-      # 10AM JST every week day
-      - schedule: cron(0 1 ? * MON-FRI *)
-        command: bin/chaos
+    scheduled_tasks: []
```
essa, line 29: You are removing all scheduled task now?

essa (approved):
@davidsiaw 
LGTM if you are removing scheduled tasks

barcelona.yml
davidsiaw: No, sorry let me remove that

PR #797, opened 2023-08-31
Recreate stack instead of erroring

davidsiaw:
Barcelona so far has simply thrown errors if it encounters a stack in `ROLLBACK_COMPLETE` since it is a dead state. However, this means by re-running deployment, the same error comes up again and again ad infinitum. A developer needs to go into AWS to delete the stack. 

This is toil.

This PR attempts to eliminate this toil by automatically deleting rollbacked stacks and re-creating them, allowing devs to simply re-run the deploy again and be able to expect a result.


lib/cloud_formation/executor.rb
```diff
@@ -107,14 +107,26 @@ def create_or_update
...
     end
 
+    def try_recreate
+      puts 'Attempting to re-create stack from ROLLBACK_COMPLETE'
+      delete
+
+      loop do
+        break if stack_status.nil? || stack_status == 'DELETE_COMPLETE'
+        puts "Waiting for stack '#{stack.name}' to delete..."
+        sleep 10
```
davidsiaw, line 123: This runs in a worker so we can take our time

app/models/backend/ecs/v2/adapter.rb
```diff
@@ -49,16 +49,19 @@ def cf_executor
...
 
+    def service_resource_id
+      cf_executor.resource_ids["ECSService"]
+    end
+
     def ecs_service
       @ecs_service ||= begin
                          executor = cf_executor
-                         if executor.stack_status.nil?
+                         if executor.stack_status.nil? || service_resource_id.nil?
```
davidsiaw, line 59: sometimes the deployment fails and rolls back, this causes service to be nil.

degikko:
@matafc can you help us review this PR, please?

davidsiaw:
Really just trying to resolve some pain that devs are going through in the meantime

lib/cloud_formation/executor.rb
```diff
@@ -107,14 +107,26 @@ def create_or_update
...
-        # The only way to solve is to delete and re-create the stack
-        raise CannotUpdateRolledbackStackException
+        try_recreate
       else
         raise UpdateInProgressException
       end
     end
 
+    def try_recreate
+      puts 'Attempting to re-create stack from ROLLBACK_COMPLETE'
```
matafc, line 117: Can we use `Rails.logger` for emitting log statements here and the other places?

davidsiaw: Oh yeah good point. WIll do

matafc (changes requested):
Overall change looks good. Requesting some logging changes. 
Also, I am not sure if it is feasible to test this on a running barcelona instance instead of just spec test, but if you are able to can you share results? I assume the difficulty with that would be to recreate the stack error reliably, if that is not possible right now that is fine.

matafc (approved):

PR #805, opened 2024-02-27
Datadog security

essa:
This PR adds command and files for enabling Datadog CSW feature to datadog plugin.

The config changes needed is described here.
- https://www.notion.so/Datadog-as-IDS-IPS-and-FIM-109f205717b34e798bf4a69aab58997b
- https://docs.datadoghq.com/security/cloud_security_management/setup/csm_cloud_workload_security/agent/linux/

### Background

Barcelona plugin hooks creating cloud-init config file and insert additional commands and files to it.

PR #806, opened 2024-02-28
Add priority to plugin hook

essa:
Continue from #805 

The Datadog agent must be invoked after the instance has been set up, otherwise it will alert about setup activity, such as installing a yum package.

This PR adds a priority to `Plugin` and updates the calling order of plugin hooks so that we can be sure that datadog is installed and invoked last.

This plugin should be registered with max priority in the district.
```
$ bcn district put-plugin -a api_key=8e53.... -a hook_priority=10 ec-staging datadog
```

And it also updates the instance type of bastion to 't3.small'. 't3.micro' was too small for all the plugin and sometimes the cloud-init fails.

### DIFF
--- a/db/migrate/20240228032657_add_hook_order_to_plugin.rb
+++ b/db/migrate/20240228032657_add_hook_order_to_plugin.rb
@@ -1,5 +1,0 @@
-class AddHookOrderToPlugin < ActiveRecord::Migration[5.2]
-  def change
-    add_column :plugins, :hook_priority, :integer, :default => 10
-  end
-end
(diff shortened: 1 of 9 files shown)


essa:
@showwin 
Sorry, hold the review please.
I found one issue to fix.

db/migrate/20240228032657_add_hook_order_to_plugin.rb
```diff
@@ -0,0 +1,5 @@
+class AddHookOrderToPlugin < ActiveRecord::Migration[5.2]
+  def change
+    add_column :plugins, :hook_priority, :integer, :default => 10
```
showwin, line 3: I think it's more natural to use `0` as the default value, but 10 is also good.
For example, comparing the following to cases, "b" looks straightforward to me. 
a.`9` is earlier than the default and `11` is later than the default.
b. `-1` is earlier than the default and `1` is later than the default.

showwin (approved):
LGTM 👍

essa:
@showwin 
I moved the `hook_priority` from a model attribute to a plugin attribute.
To add the priority by `bcn` cli, it must be a plugin attribute.

And I also make the default value to zero as your suggestion.

Could you review it again? 🙏

PR #807, opened 2024-02-29
Remove docker feature from datadog agent on bastion 

essa:
continues from #805, #806.

The configuration of datadog agent includes setting it for docker. It fails on bastion servers because bastion doesn't have docker installed.

Installation of datadog agent must be different on servers without docker. This PR fixes it.


degikko:
@showwin can you help us review this PR, please?

showwin (approved):
👍

PR #808, opened 2024-03-13
Add a switch for enabling datadog security

essa:
Continues from #807 
Datadog CWS was expensive so we want to install it to only the district really needed.
This PR adds a switch for it.

```
$ bcn district put-plugin -a api_key=8e53.... -a hook_priority=10 -a cws=enabled ec-staging datadog
```
CWS is enabled only when `cws=enabled` is specified.

And it also change the tag `role` for bastion.


degikko:
@showwin can you help us review this PR, please?

showwin (approved):
LGTM 👍

PR #811, opened 2024-05-10
increase the startup timeout for the run pack

davidsiaw:
When the ECS_IMAGE_PULL_BEHAVIOR is default and the image is large, starting a console can take more time than 30 seconds. This will lead to failure to start the container. This just loosens that startup timeout.

How to check that this is working?

Only the logic has changed. The manual test riggings no longer work so need to be updated. But the test mocks the process lister and checks that the process does what it does so `go test` needs to be good enough for now. I added more modern go-modules so you can use it if you have go installed locally. Otherwise brew install go.

1. cd dockerfiles/run-pack
2. `go build`
3. `go test`
4. should pass.


degikko:
@ochko can you help us review this PR, please?

ochko (approved):
Looks good. The test pass.

PR #816, opened 2024-08-26
Replace dots in review app subdomains

sulami:
While dots are valid, they don't work with our SSL certificate. I've already adapted the hats CI workflow to expect this change.


showwin (approved):
Thanks!

PR #820, opened 2024-11-05
Increase max body size to handle POST requests larger than 10MB

showwin:
## Context
A payload in some POST requests could be higher than 10MB, especially when users upload images or files.
Ideally, I want to increase the value to 50MB, but at the same time, I don't want to increase the memory allocation for the Nginx container to minimize the impact. So, I feel 30MB is the maximum acceptable value for 128MB memory allocation.

### DIFF
--- a/dockerfiles/reverse-proxy/templates/nginx.conf.erb
+++ b/dockerfiles/reverse-proxy/templates/nginx.conf.erb
@@ -20,7 +20,7 @@ http {
   tcp_nodelay on;
 
   # allow large file upload
-  client_max_body_size 10m;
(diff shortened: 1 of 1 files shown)


Daivasmara (approved):
LGTM, putting this discussion here for more added context: https://degica.slack.com/archives/CTG3QN2C9/p1730774247557969

showwin:
@Daivasmara I intentionally didn't add the Slack link since this is a public repo 😄 (It's safe though)

Daivasmara:
@showwin ahhhhh I forgot this is public sorry 😵

PR #823, opened 2025-02-03
Revert "[DOCKERFILES][REVERSE_PROXY] Log: $upstream_queue_time"

Daivasmara:
This reverts commit 7246164d3cd999ce846d6b4a273dd896e2122afe.

Failed CI: https://github.com/degica/barcelona/actions/runs/13104654021

somehow `$upstream_queue_time` variable is not available to use even though it's available [here in the docs](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_queue_time)

Screenshot from 2025-02-03 09-32-59

I'm still figuring out why and how to fix this, but in the meantime I prefer to revert this first

### DIFF
--- a/dockerfiles/reverse-proxy/render_template.rb
+++ b/dockerfiles/reverse-proxy/render_template.rb
@@ -68,7 +68,7 @@ def default_log_format
     '$proxy_protocol_addr - [$time_local] '
     '"$request" $status $body_bytes_sent '
     '"$http_referer" "$http_user_agent"'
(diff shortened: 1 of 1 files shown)


showwin (approved):
It might be because of the Nginx version?

Daivasmara:
> It might be because of the Nginx version?

@showwin if I understood the docs correctly, this variable came out at 1.13.9 and we're at 1.26, so should be non issue 🤔

